PR #11, opened 2018-05-22
CY-353 expose kerberos env validation

dusking:


cloudify/utils.py
```diff
@@ -385,6 +386,12 @@ def _shlex_split(command):
     return list(lex)
 
 
+def is_kerberos_env():
+    if os.path.exsts('/etc/krb5.conf') and find_executable('klist'):
```
tehasdf: exists

dusking: Oops

PR #19, opened 2018-05-25
CY-199 amqp_client connect: catch exceptions from making params

tehasdf:
Iterating _connection_params loads the params from files, so it might
well throw exceptions. Those must be handled, and popped up to the
calling thread, otherwise it'll just hang.

This change generalizes the ._error to also include those errors, and
pass them up as before.


tehasdf:
example error is DaemonNotFound when we pass a nonexistent name

PR #29, opened 2018-05-29
Handle old rest-client versions

tehasdf:
This can happen eg in the 2_1 tests in cloudify-manager, where we install
an old version of rest-client

### DIFF
--- a/cloudify/cluster.py
+++ b/cloudify/cluster.py
@@ -18,14 +18,26 @@ import json
 import pika
 import types
 import requests
+import warnings
 import itertools
 
 from cloudify import constants
 
 from cloudify_rest_client import CloudifyClient
 from cloudify_rest_client.client import HTTPClient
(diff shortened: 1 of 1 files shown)


cloudify/cluster.py
```diff
@@ -18,14 +18,26 @@
...
+from cloudify_rest_client.exceptions import CloudifyClientError
+
+
+try:
+    from cloudify_rest_client.exceptions import NotClusterMaster
+except ImportError:
+    warnings.warn('cloudify-rest-client does not support clustering - '
+                  'failover will be unavailable')
+
+    class NotClusterMaster(Exception):
```
mcouthon: Let's maybe inherit from `CloudifyClientError` here?

mcouthon:
Merging this because we need to progress with manager unittests, but this is probably not a good long term solution. We should revisit this and think about it more deeply.

PR #30, opened 2018-05-29
amqp_client: forward errors to the calling thread

tehasdf:
Implement a `publish` method on the `AMQPConnection` that makes
sure errors while publishing are bubbled up, and use that method
in handlers.

### DIFF
--- a/cloudify/amqp_client.py
+++ b/cloudify/amqp_client.py
@@ -303,7 +303,7 @@ class AMQPConnection(object):
             raise RuntimeError(
                 'Cannot wait when sending from the connection thread')
 
-        # the message will likely be sent from another thread (the .consume
+        # the message is going to be sent from another thread (the .consume
         # thread). If an error happens there, we must have a way to get it
         # back out, so we pass a Queue together with the message, that will
         # contain either an exception instance, or None
--- a/cloudify/amqp_client.py
+++ b/cloudify/amqp_client.py
@@ -288,6 +288,14 @@ class AMQPConnection(object):
                      If true, an exception will be raised if the message
                      cannot be sent.
         """
+        if wait and self._consumer_thread \
+                and self._consumer_thread is threading.current_thread():
+            # when sending from the connection thread, we can't wait because
+            # then we wouldn't allow the actual send loop (._process_publish)
+            # to run, because we'd block on the err_queue here
+            raise RuntimeError(
+                'Cannot wait when sending from the connection thread')
+
         # the message will likely be sent from another thread (the .consume
         # thread). If an error happens there, we must have a way to get it
         # back out, so we pass a Queue together with the message, that will
--- a/cloudify/amqp_client.py
+++ b/cloudify/amqp_client.py
@@ -181,7 +181,7 @@ class AMQPConnection(object):
 
         out_channel = self._pika_connection.channel()
         for handler in self._handlers:
-            handler.register(self._pika_connection, self._publish_queue)
+            handler.register(self)
             logger.info('Registered handler for {0} [{1}]'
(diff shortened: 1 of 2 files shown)


tehasdf:
You might think this is shit, but [no, it's necessary](https://www.youtube.com/watch?v=6ixvpLCdqkA)

cloudify/amqp_client.py
```diff
@@ -263,7 +276,46 @@ def close(self, wait=True):
...
+        :param message: Kwargs for the pika basic_publish call. Should at
+                        least contain the "body" and "exchange" keys, and
+                        it might contain other keys such as "routing_key"
+                        or "properties"
+        :param wait: Whether to wait for the message to actually be sent.
+                     If true, an exception will be raised if the message
+                     cannot be sent.
+        """
+        if wait and self._consumer_thread \
+                and self._consumer_thread is threading.current_thread():
```
mcouthon: If this behavior is broken, why is it allowed in the first place?

cloudify/amqp_client.py
```diff
@@ -263,7 +276,46 @@ def close(self, wait=True):
...
+            # when sending from the connection thread, we can't wait because
+            # then we wouldn't allow the actual send loop (._process_publish)
+            # to run, because we'd block on the err_queue here
+            raise RuntimeError(
+                'Cannot wait when sending from the connection thread')
+
+        # the message will likely be sent from another thread (the .consume
```
mcouthon: What do you mean "likely"? Isn't it always the case?

cloudify/amqp_client.py
```diff
@@ -263,7 +276,46 @@ def close(self, wait=True):
...
+        # thread). If an error happens there, we must have a way to get it
+        # back out, so we pass a Queue together with the message, that will
+        # contain either an exception instance, or None
+        err_queue = Queue.Queue() if wait else None
+        envelope = {
+            'message': message,
+            'err_queue': err_queue
+        }
+        self._publish_queue.put(envelope)
+        err = err_queue.get()
```
mcouthon: This will throw if `err_queue` is `None`

mcouthon: Also, do we want it to block forever, or do we want a timeout here? Can there be an unhandled scenario on the other side, that will cause this thread to block forever?

cloudify/amqp_client.py
```diff
@@ -263,7 +276,46 @@ def close(self, wait=True):
...
+        if isinstance(err, Exception):
+            raise err
```
mcouthon: Do we want to handle the case where `err` is something other than an Exception?

cloudify/amqp_client.py
tehasdf: it's not, this line disallows it :P
other than that, we have no way to stop callbacks from calling `.publish`

cloudify/amqp_client.py
tehasdf: what else could it be? currently it's always None in the success case

cloudify/amqp_client.py
```diff
@@ -263,7 +276,47 @@ def close(self, wait=True):
...
+        # back out, so we pass a Queue together with the message, that will
+        # contain either an exception instance, or None
+        err_queue = Queue.Queue() if wait else None
+        envelope = {
+            'message': message,
+            'err_queue': err_queue
+        }
+        self._publish_queue.put(envelope)
+        if err_queue:
+            err = err_queue.get(timeout=timeout)
```
tehasdf: and yeah, about the timeout, nothing uses it atm, but perhaps some thing actually should...

PR #31, opened 2018-05-29
amqp_client: ReqResp declares the queue/exchange it sends to

tehasdf:
Also split TaskConsumer into the consumer base and a mixin that
adds the threadpool processing (turned out to not be required but
i suppose still nice).

### DIFF
--- a/cloudify/amqp_client.py
+++ b/cloudify/amqp_client.py
@@ -273,10 +273,10 @@ def agent_queue_name(queue, routing_key):
 class ProcessMixin(object):
     """Implement processing incoming messages using a threadpool"""
 
-    def __init__(self, *args, **kwargs):
-        super(ProcessMixin, self).__init__(*args, **kwargs)
-        self.threadpool_size = kwargs.get('threadpool_size', 5)
+    def __init__(self, threadpool_size=5, *args, **kwargs):
+        self.threadpool_size = threadpool_size
         self._sem = threading.Semaphore(self.threadpool_size)
+        super(ProcessMixin, self).__init__(*args, **kwargs)
 
     def process(self, channel, method, properties, body):
         try:
--- a/cloudify/amqp_client.py
+++ b/cloudify/amqp_client.py
@@ -266,34 +266,17 @@ class AMQPConnection(object):
             handler.register(self._pika_connection, self._publish_queue)
 
 
-class TaskConsumer(object):
-    routing_key = ''
+def agent_queue_name(queue, routing_key):
+    return '{0}_{1}'.format(queue, routing_key)
 
-    def __init__(self, queue, threadpool_size=5):
-        self.threadpool_size = threadpool_size
-        self.exchange = queue
-        self.queue = '{0}_{1}'.format(queue, self.routing_key)
-        self._sem = threading.Semaphore(threadpool_size)
-        self._output_queue = None
-        self.in_channel = None
 
-    def register(self, connection, output_queue):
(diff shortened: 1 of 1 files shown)


cloudify/amqp_client.py
```diff
@@ -266,34 +266,17 @@ def add_handler(self, handler):
...
-                              durable=True,
-                              auto_delete=False)
-        channel.queue_bind(queue=self.queue,
-                           exchange=self.exchange,
-                           routing_key=self.routing_key)
-        channel.basic_consume(self.process, self.queue)
+    def __init__(self, threadpool_size=5, *args, **kwargs):
+        self.threadpool_size = threadpool_size
+        self._sem = threading.Semaphore(self.threadpool_size)
+        super(ProcessMixin, self).__init__(*args, **kwargs)
```
mcouthon: Let's add a comment here re: this super call, so that posterity won't try to remove it as unnecessary.

cloudify/amqp_client.py
```diff
@@ -335,6 +318,38 @@ def handle_task(self, full_task):
         raise NotImplementedError()
 
 
+class TaskConsumer(ProcessMixin):
+    routing_key = ''
+
+    def __init__(self, queue, *args, **kwargs):
```
mcouthon: Do we need to keep calling it a `queue` if it's really an exchange?

cloudify/amqp_client.py
```diff
@@ -335,6 +318,38 @@ def handle_task(self, full_task):
...
+        self.queue = agent_queue_name(queue, self.routing_key)
+        self._output_queue = None
+        self.in_channel = None
+
+    def register(self, connection, output_queue):
+        self._output_queue = output_queue
+        self.in_channel = connection.channel()
+        self._register_queue(self.in_channel)
+
+    def _register_queue(self, channel):
```
mcouthon: If `in_channel` is an attribute, might as well not pass it as an argument

cloudify/amqp_client.py
```diff
@@ -374,16 +389,24 @@ def publish(self, message, **kwargs):
...
 
-class _RequestResponseHandlerBase(TaskConsumer):
-    def __init__(self, exchange):
-        super(_RequestResponseHandlerBase, self).__init__(exchange)
-        self.queue = None
-
-    def _register_queue(self, channel):
+class _RequestResponseHandlerBase(object):
+    def __init__(self, exchange, routing_key='', *args, **kwargs):
+        super(_RequestResponseHandlerBase, self).__init__(*args, **kwargs)
```
mcouthon: Let's add a comment here re: this super call, so that posterity won't try to remove it as unnecessary.

cloudify/amqp_client.py
```diff
@@ -374,16 +389,24 @@ def publish(self, message, **kwargs):
...
+        self.exchange = exchange
+        self.routing_key = routing_key
         self.queue = uuid.uuid4().hex
-        self.in_channel.queue_declare(queue=self.queue, exclusive=True,
-                                      durable=True)
-        channel.basic_consume(self.process, self.queue)
+        self.send_queue = agent_queue_name(exchange, routing_key)
+
+    def register(self, connection, output_queue):
```
mcouthon: This looks eerily similar to `_register_queue`. Why do we need both?

geokala:
Closed: As this PR is more than six months old without having been merged. Please reopen (and rebase if necessary) if it is still relevant.

PR #37, opened 2018-06-06
Force cancel: also stop the graph execution

tehasdf:
To make it do at least as much as regular cancel


mcouthon:
This makes sure that the child thread will also quit (by raising an exception). Otherwise, doing force cancel without cancel first would result in the child thread still running.

PR #41, opened 2018-06-10
CY-336 Multiprocess logging

tehasdf:
Logs from subprocesses are sent to the main process via stdout, and
only the main process writes to the file

### DIFF
--- a/cloudify/dispatch.py
+++ b/cloudify/dispatch.py
@@ -201,8 +201,7 @@ class TaskHandler(object):
             self.run_subprocess(command_args,
                                 env=env,
                                 bufsize=1,
-                                close_fds=os.name != 'nt',
-                                cwd=dispatch_dir)
+                                close_fds=os.name != 'nt')
             with open(os.path.join(dispatch_dir, 'output.json')) as f:
                 dispatch_output = json.load(f)
             if dispatch_output['type'] == 'result':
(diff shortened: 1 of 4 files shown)


cloudify/dispatch.py
```diff
@@ -82,8 +82,44 @@
...
+
+    def write(self, data):
+        with self._lock:
+            self._f.write(data)
+            self._f.flush()
+
+
 class TaskHandler(object):
     NOTSET = object()
+    SETUP_LOGGER_LOCK = threading.Lock()
```
mcouthon: Probably don't need it anymore

cloudify/logs.py
```diff
@@ -348,42 +348,57 @@ def _publish_message(client, message, message_type, logger):
...
-    root_logger.setLevel(logging.INFO)
+    root_logger.setLevel(log_level)
     root_logger.addHandler(console_handler)
+
+    # the 'ctx' logger is for ctx.logger and will be handled by one
+    # of the PluginHandlers
+    context_logger = logging.getLogger('ctx')
+    context_logger.propagate = False
+
+    if not os.path.exists(log_dir):
```
mcouthon: Won't it break if log_dir is None? Can it be None? (it seems like it can)

cloudify/dispatch.py
```diff
@@ -82,8 +82,44 @@
     '%(asctime)s [%(name)s] %(levelname)s: %(message)s')
 
 
+class LockedFile(object):
+    SETUP_LOGGER_LOCK = threading.Lock()
+    LOGFILES = {}
+
+    @classmethod
+    def open(cls, fn):
```
mcouthon: Let's maybe add some comments here

PR #53, opened 2018-06-20
Sphinx2

opencm:


tox.ini
```diff
@@ -92,21 +77,36 @@ deps =
     {[testenv]deps}
 commands=nosetests --with-cov --cov script_runner script_runner/tests
 
+[testenv:flake8_plugins_common]
+deps =
+    flake8
+commands=flake8 cloudify --max-line-length=120
```
tyacbovi: flake8 passes with default settings, so we don't need this exception

circle.yml
```diff
@@ -27,7 +27,17 @@ test:
...
     - case $CIRCLE_NODE_INDEX in 0) tox -e flake8_script_plugin ;; esac:
         parallel: true
+    - case $CIRCLE_NODE_INDEX in 0) tox -e docs_rest_client ;; esac:
+        parallel: true
+    - case $CIRCLE_NODE_INDEX in 0) tox -e docs_plugins_common ;; esac:
+        parallel: true
+
 deployment:
+  prod:
+    branch: sphinx2
```
tyacbovi: Why to this one?

PR #54, opened 2018-06-23
Added 'type' field for mock node context

isaac-s:


mcouthon:
@isaac-s Where is this used?

isaac-s:
@mcouthon only in a unit test that I'm writing at the moment. I guess unit tests so far never really needed it... but it is needed. Am I missing anything?

PR #59, opened 2018-06-26
CY-432 Implement plugin update (via deployment update)

mcouthon:


cloudify/plugins/lifecycle.py
```diff
@@ -377,35 +377,40 @@ def node_get_state_handler(tsk):
 
 def prepare_running_agent(host_node_instance):
     tasks = []
-    install_method = utils.internal.get_install_method(
-        host_node_instance.node.properties)
-
     plugins_to_install = filter(lambda plugin: plugin['install'],
                                 host_node_instance.node.plugins_to_install)
+
+    _plugin_install_task = plugin_install_task(host_node_instance,
```
yarivgiga: better call it plugins, instead of plugin

cloudify/plugins/lifecycle.py
```diff
@@ -394,6 +394,17 @@ def prepare_running_agent(host_node_instance):
...
 
+def plugin_uninstall_task(host_node_instance, plugins_to_uninstall):
+    install_method = utils.internal.get_install_method(
+        host_node_instance.node.properties)
+    if (plugins_to_uninstall and
+            install_method != constants.AGENT_INSTALL_METHOD_NONE):
+        return host_node_instance.execute_operation(
+            'cloudify.interfaces.cloudify_agent.uninstall_plugins',
+            kwargs={'plugins': plugins_to_uninstall})
+    return None
```
yarivgiga: just to be more clear?

cloudify/plugins/workflows.py
```diff
@@ -599,3 +605,48 @@ def _add_validate_to_task_graph(graph, hosts, current_amqp, manager_ip=None,
...
+    # {'add': (NODE_ID, PLUGIN_DICT), 'remove': (NODE_ID, PLUGIN_DICT)}]
+    # So we filter out only those dicts that have the relevant action
+    plugins_to_handle = [p[action] for p in plugins_list if p.get(action)]
+
+    # The list might contain duplicates, and it's organized in the following
+    # way: [(node_id, plugin_dict), (node_id, plugin_to_handle), ...] so
+    # we reorganize it into: {node_id: [list_of_plugins], ...}
+    node_to_plugins_map = {}
+    for node, plugin in plugins_to_handle:
+        node_list = node_to_plugins_map.setdefault(node, [])
```
yarivgiga: would have done it in one line:
node_to_plugins_map.setdefault(node, set()).add(plugin)

cloudify/plugins/lifecycle.py
```diff
@@ -377,35 +377,52 @@ def node_get_state_handler(tsk):
...
 
-    plugins_to_install = filter(lambda plugin: plugin['install'],
-                                host_node_instance.node.plugins_to_install)
+
+def plugin_install_task(host_node_instance, plugins_to_install):
```
yarivgiga: better "plugins" in plural

cloudify/plugins/lifecycle.py
mcouthon: Yep. It feels more explicit

cloudify/plugins/workflows.py
mcouthon: But I'm doing another if check after this

cloudify/plugins/lifecycle.py
mcouthon: Fair enough

cloudify/plugins/workflows.py
yarivgiga: this is why i suggested using a set and not a list... to save the if

PR #74, opened 2018-08-02
CY-539-fix-blueprint-tar-permissions

idob-cloudify:


cloudify_rest_client/utils.py
```diff
@@ -28,10 +28,14 @@ def tar_file(file_to_tar, destination_dir, tar_name=''):
     :param tar_name: optional tar name.
     :return:
     """
+    def _reset_tarinfo(tarinfo):
+        tarinfo.mode = 0755
```
tehasdf: might as well use `0o755` syntax for py2+3-compat

isaac-s: - What about file ownership? can / should we guarantee `cfyuser:cfyuser`? Maybe `cfyuser:adm`?
- Not sure that `0755` is the right thing to do (depends on the file ownership we choose to go with)

tehasdf: it is going to be cfyuser guaranteed, because that's the user that saves the file (restservice)

isaac-s: @tehasdf and the group? will be the primary group of `cfyuser`, right?

tehasdf: i believe so

tehasdf:
@isaac-s I've dismissed your review, because this PR is no longer setting 0755, it's now setting o+r

PR #75, opened 2018-08-03
dispatch.py with: python 2.6-compatible syntax

tehasdf:


cloudify/dispatch.py
```diff
@@ -403,8 +403,9 @@ def handle(self):
         if not self.cloudify_context.get('no_ctx_kwarg'):
             kwargs['ctx'] = ctx
 
-        with state.current_ctx.push(ctx, kwargs), self._amqp_client():
-            result = self._run_operation_func(ctx, kwargs)
+        with state.current_ctx.push(ctx, kwargs):
```
geokala: Can we please put a comment on this stating what it should be but isn't, and why (python2.6)?

tehasdf: done

PR #98, opened 2018-09-25
kerberos-improvements

idob-cloudify:


cloudify_rest_client/client.py
```diff
@@ -152,7 +152,9 @@ def verify_response_status(self, response, expected_code=200):
     def _do_request(self, requests_method, request_url, body, params, headers,
                     expected_status_code, stream, verify, timeout):
         auth = None
-        if self.has_kerberos():
+        if self.has_kerberos() and \
+            CLOUDIFY_TOKEN_AUTHENTICATION_HEADER not in self.headers and \
```
tehasdf: thats a long if, maybe factor out and `if self.has_kerberos() and not self.has_cloudify_auth()`?

idob-cloudify: updated.

PR #109, opened 2018-10-09
CY-680: adding support for blueprint import

tyacbovi:
- Added recognition of blueprint catalog import.
- Removed not needed functions in the abstract resolver.
- Added the ability to parse a blueprint and receive the merged blueprint, this is for being able to import a blueprint from the catalog.

- Add fixing CY-689: description from imported blueprint should not try to overwrite the existing one

### DIFF
--- a/dsl_parser/elements/imports.py
+++ b/dsl_parser/elements/imports.py
@@ -48,6 +48,10 @@ MERGEABLE_FROM_DSL_VERSION_1_3 = [
     constants.NODE_TEMPLATES
 ]
 
+DONT_OVERWRITE = set([
+    constants.DESCRIPTION
+])
+
 IGNORE = set([
     constants.DSL_DEFINITIONS,
     constants.IMPORTS,
@@ -263,6 +267,8 @@ def _merge_parsed_into_combined(combined_parsed_dsl_holder,
             pass
         elif key_holder.value not in combined_parsed_dsl_holder:
             combined_parsed_dsl_holder.value[key_holder] = value_holder
+        elif key_holder.value in DONT_OVERWRITE:
+            pass
         elif key_holder.value in merge_no_override:
             _, to_dict = combined_parsed_dsl_holder.get_item(key_holder.value)
             _merge_into_dict_or_throw_on_duplicate(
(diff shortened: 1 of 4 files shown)


dsl_parser/parser.py
```diff
@@ -54,13 +54,40 @@ def _parse(dsl_string,
...
+            'resource_base': resource_base,
+            'validate_version': validate_version
+        },
+        element_cls=blueprint.Blueprint)
+
+    functions.validate_functions(plan)
+    return plan
+
+
+def _resolve_blueprint_imports(dsl_location,
```
mcouthon: Make public if using outside this module

dsl_parser/import_resolver/abstract_import_resolver.py
```diff
@@ -39,14 +39,8 @@ class AbstractImportResolver(object):
...
     @abc.abstractmethod
-    def resolve(self, import_url):
-        raise NotImplementedError
-
     def fetch_import(self, import_url):
-        url_parts = import_url.split(':')
-        if url_parts[0] in ['http', 'https', 'ftp', 'file']:
-            return self.resolve(import_url)
-        return read_import(import_url)
+        raise NotImplementedError
```
tehasdf: i think we used to, or still do, provide a way for the user to pass their own import resolvers, right? if so, the abstract import resolver is public api, and we should keep some backwards compat here: if the user used to implement just `resolve`, because that was all that was needed, this change will break their resolver... can we do it in a compatible way?

dsl_parser/elements/imports.py
```diff
@@ -192,6 +197,9 @@ def location(value):
     imports_graph = ImportsGraph()
     imports_graph.add(location(dsl_location), parsed_dsl_holder)
 
+    def _is_parsed_resource(imported_resource):
+        return isinstance(imported_resource, Holder)
```
tehasdf: so the new resolver in cloudify-cosmo/cloudify-manager#1486 returns a Holder?

tyacbovi: Yes, the yaml

PR #110, opened 2018-10-10
TaskDispatcher: for mgmtworker queue, use the mgmtworker user

tehasdf:
Don't send messages to the mgmtworker using the tenant user on
vhost /, use the mgmtworker (config) user instead

### DIFF
--- a/cloudify/workflows/workflow_context.py
+++ b/cloudify/workflows/workflow_context.py
@@ -1202,7 +1202,7 @@ class _TaskDispatcher(object):
             client = amqp_client.get_client(
                 amqp_user=tenant['rabbitmq_username'],
                 amqp_pass=tenant['rabbitmq_password'],
-                amqp_vhost=task['tenant']['rabbitmq_vhost']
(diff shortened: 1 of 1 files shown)


cloudify/workflows/workflow_context.py
```diff
@@ -1192,21 +1192,18 @@ def make_subtask(self, tenant, target, queue, *args, **kwargs):
...
-            amqp_pass=tenant['rabbitmq_password'],
-            amqp_vhost=self._get_vhost(task)
-        )
+        if task['queue'] == MGMTWORKER_QUEUE:
+            client = amqp_client.get_client()
+        else:
+            client = amqp_client.get_client(
+                amqp_user=tenant['rabbitmq_username'],
+                amqp_pass=tenant['rabbitmq_password'],
+                amqp_vhost=task['tenant']['rabbitmq_vhost']
```
inbala: There is no tenant['rabbitmq_vhost']?

PR #112, opened 2018-10-16
disable chunks in blueprint upload when working in kerberos

idob-cloudify:


cloudify_rest_client/bytes_stream_utils.py
```diff
@@ -19,6 +19,30 @@
...
+                             client=None):
+    """
+    Read file data in kerberos mode or split into chunks,
+    :param file_path: Local path of the file to be transferred
+    :param buffer_size: Size of the buffer
+    :param progress_callback: Callback function - can be used to print progress
+    :return: Generator object
+    """
+
+    if client and client.has_kerberos() and not client.has_auth_header():
```
tyacbovi: Make this a bool input for this function, so the logic if to send in chunks will be in the user's side and can be reused elsewhere

cloudify_rest_client/bytes_stream_utils.py
```diff
@@ -19,6 +19,30 @@
...
 
+def request_data_file_stream(file_path,
+                             progress_callback=None,
+                             client=None):
+    """
+    Read file data in kerberos mode or split into chunks,
+    :param file_path: Local path of the file to be transferred
+    :param buffer_size: Size of the buffer
+    :param progress_callback: Callback function - can be used to print progress
+    :return: Generator object
```
tyacbovi: Or will return the entire file

cloudify_rest_client/bytes_stream_utils.py
```diff
@@ -19,6 +19,30 @@
 DEFAULT_BUFFER_SIZE = 8192
 
 
+def request_data_file_stream(file_path,
+                             progress_callback=None,
+                             client=None):
+    """
+    Read file data in kerberos mode or split into chunks,
+    :param file_path: Local path of the file to be transferred
+    :param buffer_size: Size of the buffer
```
tyacbovi: maybe to allow this also so the user can also configure that option

PR #121, opened 2018-10-29
Cy 634 add cloudify cluster support for external db implementations

barucoh:


cloudify_rest_client/exceptions.py
```diff
@@ -260,4 +270,5 @@ class DeploymentPluginNotFound(CloudifyClientError):
         PluginInstallationTimeout,
         NotClusterMaster,
         RemovedFromCluster,
-        DeploymentPluginNotFound]])
+        DeploymentPluginNotFound,
+        IncompatibleClusterArchitectureException]])
```
mcouthon: All the others seem to be Errors, so maybe this one should be as well

cloudify_rest_client/exceptions.py
```diff
@@ -234,6 +234,16 @@ class DeploymentPluginNotFound(CloudifyClientError):
     ERROR_CODE = 'deployment_plugin_not_found'
 
 
+class IncompatibleClusterArchitectureException(CloudifyClientError):
```
mcouthon: Where is this class actually used?

cloudify_rest_client/exceptions.py
```diff
@@ -242,6 +242,16 @@ class DeploymentPluginNotFound(CloudifyClientError):
     ERROR_CODE = 'deployment_plugin_not_found'
 
 
+class IncompatibleClusterArchitectureError(CloudifyClientError):
+    """
+    Raise when a cluster node with architecture X is trying to join a cluster
+    with architecture Y
+
+    I.E - Master is all-in-one and slave has an external database
```
tehasdf: btw `IE` = "id est" = "that is"
I think you meant "EG" = "for example"

cloudify_rest_client/cluster.py
```diff
@@ -163,14 +167,21 @@ def add(self, host_ip, node_name):
...
 
+        database_ip is used to validate architecture compatibility of the
+        cluster nodes
+
         :param host_ip: the externally-visible IP of the current node
         :param node_name: the name of this node used internally in the cluster
         :return: representation of the node that will join the cluster
         :rtype: ClusterNode
         """
+        response = self.api.get('/rest/config', filter='postgresql_host')
```
tehasdf: it's weird that we ask the node for its postgres host, and then send it back to it, no? i'd prefer for the node to figure it out itself, because this request means that joining is now 3 requests instead of 2

cloudify_rest_client/exceptions.py
tehasdf: teacher lied to you lol :) https://en.wiktionary.org/wiki/i.e.

cloudify_rest_client/cluster.py
barucoh: Fixed, moved validations to the joining node's side rather than the Master node's

cloudify_rest_client/cluster.py
```diff
@@ -163,6 +167,9 @@ def add(self, host_ip, node_name):
         Use the credentials returned as part of the ClusterNode structure
         when joining the cluster.
 
+        database_ip is used to validate architecture compatibility of the
```
tehasdf: comment is now useless

barucoh: Fixed

PR #122, opened 2018-10-30
CY-757: added post-operation events

isaac-s:


cloudify/plugins/lifecycle.py
```diff
@@ -192,26 +192,32 @@ def install_node_instance_subgraph(instance, graph, **kwargs):
...
                  instance.set_state('creating')),
         instance.execute_operation('cloudify.interfaces.lifecycle.create'),
-        instance.set_state('created'),
+        forkjoin(instance.send_event('Node instance created'),
+                 instance.set_state('created')),
+        instance.send_event('Pre-configuring relationships'),
         _relationships_operations(
             subgraph,
             instance,
             'cloudify.interfaces.relationship_lifecycle.preconfigure'
```
geokala: Should there be an event here to announce that we are pre-configuring the relationships in the same way as there is on the post-configuring one?

isaac-s: .... But it's there. What am I missing?

PR #127, opened 2018-11-09
CY-781: stdout/stderr printed to ctx logger by default

isaac-s:
To do this, I restored much of the code from script plugin 1.5.1.

### DIFF
--- a/script_runner/tasks.py
+++ b/script_runner/tasks.py
@@ -265,11 +265,15 @@ def execute(script_path, ctx, process):
     else:
         ctx.logger.debug("Context proxy closed")
 
-    for t, name in [(stdout_consumer, 'stdout'), (stderr_consumer, 'stderr')]:
-        if t:
+    for consumer, name in [(stdout_consumer, 'stdout'),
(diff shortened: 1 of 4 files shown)


script_runner/tasks.py
```diff
@@ -284,6 +262,14 @@ def execute(script_path, ctx, process):
...
     except Exception:
         ctx.logger.warning('Failed closing context proxy', exc_info=True)
+    else:
+        ctx.logger.debug("Context proxy closed")
+
+    for t, name in [(stdout_consumer, 'stdout'), (stderr_consumer, 'stderr')]:
+        if t:
+            ctx.logger.debug('Joining consumer thread for %s', name)
+            t.join()
+            ctx.logger.debug('Consumer thread for %s ended', name)
```
EarthmanT: For readability, can you change `t` to something like `consumer`?

isaac-s: Yes, one sec.

isaac-s: Done.

PR #137, opened 2018-11-20
CY-783 Select group member instances in scale down

geokala:


cloudify/plugins/workflows.py
```diff
@@ -76,6 +78,132 @@ def auto_heal_reinstall_node_subgraph(
                                        ignore_failure=ignore_failure)
 
 
+def get_groups_with_members(ctx):
```
tehasdf: document what this returns, eg a docstring saying `{scaling group id: [node instance ids]}`

cloudify/plugins/workflows.py
```diff
@@ -76,6 +78,132 @@ def auto_heal_reinstall_node_subgraph(
...
+    for instance in ctx.node_instances:
+        scaling_group_ids = [sg.get('id') for sg in instance.scaling_groups]
+        for sg_id in scaling_group_ids:
+            if sg_id not in groups_members:
+                groups_members[sg_id] = []
+            groups_members[sg_id].append(instance.id)
+    return groups_members
+
+
+def _check_for_too_many_exclusions(exclude_instances, available_instances,
```
tehasdf: please docstring what does this check, and what are the args (string - ids? or instances, or dicts)

cloudify/plugins/workflows.py
```diff
@@ -76,6 +78,132 @@ def auto_heal_reinstall_node_subgraph(
...
+                                   delta, groups_members):
+    if groups_members:
+        excluded_groups = set()
+        for inst in exclude_instances:
+            for group, members in groups_members.items():
+                if inst in members or inst == group:
```
tehasdf: inst can be both a group or a node instance?

tehasdf: oh, or this is for nested groups?

cloudify/plugins/workflows.py
```diff
@@ -76,6 +78,132 @@ def auto_heal_reinstall_node_subgraph(
...
+        return 'Available instances were: {instances}'.format(
+            instances=', '.join(available_instances),
+        )
+
+
+def validate_inclusions_and_exclusions(include_instances, exclude_instances,
+                                       available_instances,
+                                       delta, scale_compute,
+                                       groups_members=None):
+    if include_instances or exclude_instances:
```
tehasdf: reverse the condition and we can dedent the whole function

PR #145, opened 2018-11-29
CY-746 Add ability to get_input nested property

davidginzbourg:
`get_input` now behaves as `get_property` but also supports the old
syntax.

This enables acquiring a nested property of a provided input, in the
same way that get_property does.

### DIFF
--- a/dsl_parser/exceptions.py
+++ b/dsl_parser/exceptions.py
@@ -19,22 +19,29 @@ class MissingRequiredInputError(Exception):
     An error raised when a deployment is created and a required input
     was not specified on its creation.
     """
-    def __init__(self, *args, **kwargs):
-        super(MissingRequiredInputError, self).__init__(*args, **kwargs)
+    pass
 
 
 class UnknownInputError(Exception):
     """
     An error raised when an unknown input is specified on deployment creation.
     """
-    def __init__(self, *args, **kwargs):
-        super(UnknownInputError, self).__init__(*args, **kwargs)
+    pass
+
+
+class InputEvaluationError(Exception):
+    """
+    An error raised when the provided input cannot be evaluated (e.g. when
+    it's missing an attribute that has been required)
+    """
+    pass
 
 
 class FunctionEvaluationError(Exception):
     """
     An error raised when an intrinsic function was unable to get evaluated.
     """
+
     def __init__(self, func_name, message=None):
         msg = 'Unable to evaluate {0} function'.format(func_name)
         if message:
@@ -47,8 +54,7 @@ class UnknownSecretError(Exception):
     An error raised when a deployment is created and a required secret
     does not exist.
     """
-    def __init__(self, *args, **kwargs):
-        super(UnknownSecretError, self).__init__(*args, **kwargs)
+    pass
 
 
 class UnsupportedGetSecretError(Exception):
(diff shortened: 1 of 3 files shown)


dsl_parser/exceptions.py
```diff
@@ -31,6 +31,15 @@ def __init__(self, *args, **kwargs):
         super(UnknownInputError, self).__init__(*args, **kwargs)
 
 
+class InputEvaluationError(Exception):
+    """
+    An error raised when the provided input cannot be evaluated (e.g. when
+    it's missing an attribute that has been required)
+    """
+    def __init__(self, *args, **kwargs):
```
geokala: You don't actually need to specify this if you're not doing anything with the __init__ yourself- just remove is, put a 'pass' in and it'll fall through to the parent's __init__.

dsl_parser/functions.py
```diff
@@ -147,25 +147,83 @@ def evaluate_runtime(self, storage):
...
-        if not valid_args_type:
+        if not isinstance(args, basestring) \
+                and not (isinstance(args, list) and len(args) >= 1):
             raise ValueError(
-                "get_input function argument should be a string, "
-                "but is '{0}'.".format(args))
-        self.input_name = args
+                "Illegal argument(s) passed to {0} function. Expected either"
+                "a string or a list [input_name [, key1/index1 [,...]]] but "
+                "got {1}".format(self.name, args))
```
geokala: Maybe ', '.join([str(arg) for arg in args])?

dsl_parser/functions.py
```diff
@@ -147,25 +147,83 @@ def evaluate_runtime(self, storage):
...
+        self.input_value = args
 
     def validate(self, plan):
-        if self.input_name not in plan.inputs:
+        if isinstance(self.input_value, list):
```
geokala: This might be better with something like:
input_value = self.input_value[0] if isinstance(self.input_value, list) else self.input_value
Then a single check to see if input_value in plan.inputs, using just the last UnknownInputError?

dsl_parser/functions.py
```diff
@@ -147,25 +147,83 @@ def evaluate_runtime(self, storage):
...
+                return ''
+            s = str(attr_list[0])
+            for attr in attr_list[1:]:
+                if isinstance(attr, int):
+                    s += '[{0}]'.format(attr)
+                else:
+                    s += '.{0}'.format(attr)
+            return s
+
+        if isinstance(root, dict) and 'default' in root:
```
geokala: Do you actually get to see the dict by this point in the process? I would've expected the DSL parser to convert it beforehand?

dsl_parser/functions.py
davidginzbourg: Yeah, why not.

dsl_parser/functions.py
davidginzbourg: Sure

dsl_parser/functions.py
davidginzbourg: No, as far as I understand, the input might be something else rather than a dict.
It'll always be a dict when it's a blueprint obj, once an input is provided, it can be anything.

geokala: Yes, but by the time get_input is called won't it always have ceased being the type of dict where we care about the 'default' key?

e.g. if the input is { 'default': 'mydefault', 'description': 'Something' } then by the time your code is executed (and assuming no actual input provided), you will receive 'mydefault'.

If that's correct then this code will only trigger in cases where it probably shouldn't (when a user has entered a dict as the input with a 'default' key in it).

davidginzbourg: You're completely right. I think I added this in the early stages and never deleted it. This piece of code does indeed only get called whenever it shouldn't.
Fixed.

PR #146, opened 2018-12-01
CY-766: Avoid AMQP hang

isaac-s:


cloudify/dispatch.py
```diff
@@ -457,6 +465,7 @@ def _handle_remote_workflow(self):
         tenant = self.ctx._context['tenant'].get('original_name',
                                                  self.ctx.tenant_name)
         rest = get_rest_client(tenant=tenant)
+        # TODO: if this one throws an exception, the dispatcher hangs forever.
```
tehasdf: what does it mean "the dispatcher hangs"? from the code it looks like the execution status will never be updated to failed, but i dont see why would anything hang

PR #147, opened 2018-12-02
CY-642: improving blueprint imported list

tyacbovi:
- The imported list will contain only blueprint imports, due that there is only a need for them
- Now the imported list will come also from the imported blueprints, if local or blueprint imports

### DIFF
--- a/dsl_parser/elements/imports.py
+++ b/dsl_parser/elements/imports.py
@@ -90,7 +90,6 @@ class ImportsLoader(Element):
     }
 
     resource_base = None
-    imported = None
 
     def validate(self, **kwargs):
         imports = [i.value for i in self.children()]
@@ -114,20 +113,18 @@ class ImportsLoader(Element):
                 resources_base_path=resources_base_path)
             slash_index = blueprint_location.rfind('/')
             self.resource_base = blueprint_location[:slash_index]
-        imports_list, parsed_blueprint = _combine_imports(
+        parsed_blueprint = _combine_imports(
             parsed_dsl_holder=main_blueprint_holder,
             dsl_location=blueprint_location,
             resources_base_path=resources_base_path,
             version=version,
             resolver=resolver,
             validate_version=validate_version)
-        self.imported = imports_list
         return parsed_blueprint
 
     def calculate_provided(self, **kwargs):
         return {
-            'resource_base': self.resource_base,
-            constants.IMPORTED: self.imported
+            'resource_base': self.resource_base
         }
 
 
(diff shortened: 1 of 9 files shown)


dsl_parser/elements/imports.py
```diff
@@ -113,20 +112,18 @@ def parse(self,
...
             resolver=resolver,
             validate_version=validate_version)
-        self.imported = imports_list
         return parsed_blueprint
 
     def calculate_provided(self, **kwargs):
         return {
-            'resource_base': self.resource_base,
-            constants.IMPORTED: self.imported
+            'resource_base': self.resource_base
```
mcouthon: If we're returning a single value, can we just return it, instead of a dict?

dsl_parser/elements/imports.py
```diff
@@ -216,11 +216,13 @@ def _combine_imports(parsed_dsl_holder, dsl_location,
...
     holder_result.value = {}
-    used_imports = set()
+    blueprint_imports = set()
     for imported in ordered_imports:
         import_url, namespace = imported['import']
-        if import_url != constants.ROOT_ELEMENT_VALUE:
-            used_imports.add(import_url)
+        if (import_url != constants.ROOT_ELEMENT_VALUE and
+                utils.is_blueprint_import(import_url)):
+            blueprint_id = utils.remove_blueprint_import_prefix(import_url)
```
mcouthon: Please explain what we're doing here

dsl_parser/elements/misc.py
```diff
@@ -116,5 +116,9 @@ class Imported(Element):
 
 
 class ImportedBlueprints(Element):
+    """
+    Internal DSL element used for maintaining a list of imported blueprints,
+    this is for enabling protection against there deletion when used.
```
mcouthon: s/there/their/

dsl_parser/elements/imports.py
tyacbovi: Sadly no, this is a part of infrastructure

dsl_parser/elements/imports.py
tyacbovi: Not relevant in the following

dsl_parser/parser.py
```diff
@@ -14,8 +14,7 @@
 #    * limitations under the License.
 
 from dsl_parser import (functions,
-                        utils,
-                        constants)
+                        utils)
```
inbala: I think it can be in one line

dsl_parser/tests/test_blueprint_import.py
```diff
@@ -0,0 +1,138 @@
...
+#    * limitations under the License.
+
+from dsl_parser import constants, utils
+from dsl_parser.exceptions import DSLParsingLogicException
+from dsl_parser.tests.abstract_test_parser import AbstractTestParser
+from ..import_resolver.default_import_resolver import DefaultImportResolver
+
+
+class ResolverWithBlueprintSupport(DefaultImportResolver):
+    BLUEPRINT_PREFIX = 'blueprint:'
```
inbala: The BLUEPRINT_IMPORT const can be used instead

PR #151, opened 2018-12-04
CY-780 amqp_client: use the heartbeat value from broker_config

tehasdf:
No need for the default after all


cloudify/amqp_client.py
```diff
@@ -116,7 +114,7 @@ def _get_daemon_factory():
         from cloudify_agent.api.factory import DaemonFactory
     except ImportError:
         # Might not exist in e.g. the REST service
-        DaemonFactory = None
+        DaemonFactory = None  # NOQA
```
geokala: Why does this need noqa? It's used on the next line isn't it?

PR #157, opened 2018-12-10
CY-874 default import resolver: fallback - make configurable

tehasdf:
Falling back to the original url can now be configured from the
pctx level

### DIFF
--- a/dsl_parser/import_resolver/default_import_resolver.py
+++ b/dsl_parser/import_resolver/default_import_resolver.py
@@ -75,7 +75,7 @@ class DefaultImportResolver(AbstractImportResolver):
     be resolved, the original url will be tried anyway.
     """
 
-    def __init__(self, rules=None, fallback=True):
+    def __init__(self, rules=None, fallback=False):
         # set the rules
         self.rules = rules
         self._fallback = fallback
--- a/dsl_parser/import_resolver/default_import_resolver.py
+++ b/dsl_parser/import_resolver/default_import_resolver.py
@@ -70,11 +70,15 @@ class DefaultImportResolver(AbstractImportResolver):
 
         In case that all the resolve attempts will fail,
         a DSLParsingLogicException will be raise.
+
+    If ``fallback`` is set (the default), then in case all rules fail to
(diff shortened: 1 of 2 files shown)


dsl_parser/import_resolver/default_import_resolver.py
```diff
@@ -70,23 +70,29 @@ class DefaultImportResolver(AbstractImportResolver):
 
         In case that all the resolve attempts will fail,
         a DSLParsingLogicException will be raise.
+
+    If ``fallback`` is set (the default), then in case all rules fail to
+    be resolved, the original url will be tried anyway.
     """
 
-    def __init__(self, rules=None):
+    def __init__(self, rules=None, fallback=True):
```
isaac-s: `False` should be the default as it's more intuitive.

isaac-s: This makes the resolver configurable at the code level, but not in a way that is usable for the customer. There should be another PR to `cloudify-manager-install` with the appropriate change in `config.yaml`.

tehasdf: @isaac-s note that the restservice creates the import resolver by just doing `resolver(**kw)` with args coming from the provider context; so this is still half-accessible by the customer, by changing the provider context

isaac-s: @tehasdf customers shouldn't, and won't, modify the provider context.
Let alone the fact that we'd like to get rid of the pctx altogether.

tehasdf:
Added a `config.yaml` frontend for this in cloudify-cosmo/cloudify-manager-install#401

dsl_parser/import_resolver/default_import_resolver.py
geokala: What's the current behaviour (and is the current behaviour something that we should consider erroneous)?

tehasdf: current behaviour is like True, ie. if the resolved url returns a 404, still try the original url. We decided that's wrong and it should just be an error instead (ie. like False)

PR #162, opened 2018-12-11
CY-642:  Preventing adding a namespace to Cloudify basic types

tyacbovi:


dsl_parser/elements/imports.py
```diff
@@ -246,6 +246,34 @@ def validate_namespace(namespace):
...
+        """
+        Cloudify basic types can not be imported with namespace,
+        due to that will disrupt core Cloudify types with the namespace prefix
+        which will not allow proper functioning, like: added a prefix to
+        install workflow.
+        """
+        if cloudify_basic_types and namespace:
+            raise exceptions.DSLParsingLogicException(
+                214,
+                'Invalid import namespace {} cannot be used'
```
tehasdf: add some punctuation

dsl_parser/elements/imports.py
```diff
@@ -259,27 +287,33 @@ def build_ordered_imports_recursive(_current_parsed_dsl_holder,
...
                 if namespace:
-                    namespace = utils.generate_namespaced_value(
+                    current_namespace = utils.generate_namespaced_value(
                         context_namespace,
                         namespace)
                 else:
                     # In case a namespace was added earlier in the import
                     # chain.
-                    namespace = context_namespace
+                    current_namespace = context_namespace
```
tehasdf: i can't really follow the flow with all 3 of current_namespace, context_namespace, and namespace - are they used correctly everywhere? eg. here https://github.com/cloudify-cosmo/cloudify-common/blob/92926404569ada64b98d5a03cd8ea43ba93ce512/dsl_parser/elements/imports.py#L327

dsl_parser/elements/operation.py
```diff
@@ -295,7 +295,8 @@ def _resource_exists(location_bases, resource_name):
         # pointing to the same installed one.
         script_plugins = [plugin for plugin in plugins
                           if constants.SCRIPT_PLUGIN_NAME in plugin]
-        if not script_plugins:
+        script_plugin = script_plugins and script_plugins[0]
```
tehasdf: kinda dont like the shortcircuiting `and` here, since we use the value assuming it is the name not a bool; what if `script_plugins[0]` is false :/ (an empty string)
maybe rewrite as an if?

dsl_parser/framework/parser.py
```diff
@@ -224,26 +224,35 @@ def handle_intrinsic_function_namespace(element):
...
                 element._initial_value = utils.generate_namespaced_value(
                         leaf_namespace, element._initial_value)
             elif isinstance(element._initial_value, dict):
                 element._initial_value = \
                     handle_intrinsic_function_namespace(element._initial_value)
 
+        def should_add_element_namespace(element_value):
+            # Preventing of adding namespace prefix to cloudify
+            # basic types.
+            return not (isinstance(element_value, basestring) and
```
tehasdf: this can be not a string?

dsl_parser/rel_graph.py
```diff
@@ -828,15 +825,9 @@ def _verify_and_get_connection_type(relationship):
...
-    def search_relationship_type(type_name, types_list):
-        # Searching in the expected types while neglecting the namespace
-        # differences, because these are types.yaml basic types.
-        return any([item for item in types_list if item in type_name])
-
-    relationship_type_hierarchy = relationship[constants.TYPE_HIERARCHY]
-    return any([search_relationship_type(relationship_type, expected_types)
+def _relationship_type_hierarchy_includes_one_of(relationship, expected_types):
+    relationship_type_hierarchy = relationship['type_hierarchy']
+    return any([relationship_type in expected_types
```
tehasdf: can drop the []: `any(relationship_type in ...)`

PR #169, opened 2018-12-19
CY-875 Disallow importing from outside the resources directory

tehasdf:
If you try to import `../../../etc/passwd`, this will stop you once
you leave the resources directory (/opt/manager/resources)


geokala:
As noted in the associated jira, this is effectively unfixable at the moment. Closing.

PR #176, opened 2018-12-23
CY-844 Add resuming to builtin workflows 

tehasdf:


tehasdf:
Currently also contains the commits from #175 - until that is merged

PR #181, opened 2018-12-26
CY-900 Add `created_at` to local env blueprints

tehasdf:


cloudify/workflows/local.py
```diff
@@ -505,9 +514,11 @@ def __init__(self, storage_dir='/tmp/cloudify-workflows'):
         self._data_path = None
         self._payload_path = None
         self._blueprint_path = None
+        self.created_at = None
```
tyacbovi: It gets already initialized in the parent class, no?

tehasdf: this one doesn't call super (for some reason)

geokala: Isn't it doing that on line 509 (of the modified version)?

tehasdf: oh yeah it does now, lol

PR #183, opened 2018-12-27
CY-884 Add support for nested intrinsic function calls

davidginzbourg:
Adds the ability to call other intrinsic functions within intrinsic functions.

### DIFF
--- a/dsl_parser/functions.py
+++ b/dsl_parser/functions.py
@@ -13,8 +13,10 @@
 #    * See the License for the specific language governing permissions and
 #    * limitations under the License.
 
-import pkg_resources
 import abc
+import pkg_resources
+
+from functools import wraps
 
 from dsl_parser import (constants,
                         exceptions,
@@ -25,16 +27,22 @@ SOURCE = 'SOURCE'
 TARGET = 'TARGET'
 AVAILABLE_NODE_TARGETS = [SELF, SOURCE, TARGET]
 
+# Function eval types: Static cannot be evaluated at runtime,
+# Hybrid can be evaluated both at runtime and statically,
+# Runtime can be evaluated only at runtime.
+STATIC_FUNC, HYBRID_FUNC, RUNTIME_FUNC = range(3)
+
 TEMPLATE_FUNCTIONS = {}
 
 
-def register(fn=None, name=None):
+def register(fn=None, name=None, func_eval_type=HYBRID_FUNC):
     if fn is None:
         def partial(_fn):
-            return register(_fn, name=name)
+            return register(_fn, name, func_eval_type)
         return partial
     TEMPLATE_FUNCTIONS[name] = fn
     fn.name = name
+    fn.func_eval_type = func_eval_type
     return fn
 
 
@@ -77,6 +85,16 @@ def _convert_attribute_list_to_python_syntax_string(attr_list):
 
 
 _register_entry_point_functions()
+
+
+def _contains_legal_nested_attribute_path_items(l):
+    return all(_is_function(x) or isinstance(x, (basestring, int)) for x in l)
+
+
+def _is_legal_nested_attribute_path(l):
+    return isinstance(l, list) \
+           and len(l) >= 2 \
+           and _contains_legal_nested_attribute_path_items(l)
 
 
 class RuntimeEvaluationStorage(object):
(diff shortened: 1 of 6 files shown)


dsl_parser/functions.py
```diff
@@ -79,6 +85,19 @@ def _convert_attribute_list_to_python_syntax_string(attr_list):
 _register_entry_point_functions()
 
 
+def _contains_legal_nested_attribute_path_items(l):
+    return all(map(
+        lambda x:
+        _is_function(x) or isinstance(x, (basestring, int)),
+        l))
```
tehasdf: can be written nicer as `return all(_is_function(x) or isinstance(x, (basestring, int)) for x in l)` without the map+lambda

dsl_parser/functions.py
```diff
@@ -243,16 +285,20 @@ def __init__(self, args, **kwargs):
...
+                '[, nested-property-1\\function, ... ] '
+                'but got: {1}.'.format(self.name, args))
         self.node_name = args[0]
         self.property_path = args[1:]
 
     def validate(self, plan):
-        self.evaluate(plan)
+        # Try to evaluate only when the arguments don't contain functions
+        args = [self.node_name] + self.property_path
+        if all(map(lambda x: not isinstance(x, dict), args)):
```
tehasdf: i think `if all(not isinstance(x, dict) for x in args)` is nicer, or maybe `if any(isinstance(x, dict) for x in args): return`

tehasdf: ditto for all usages of map+lambda, it's not a big difference but in general i'd prefer generator expressions like that because 1) i think it's more readable (less cruft from `lambda`), 2) it's slightly faster (map does a whole function call which is slow in python and a genexp just executes the expression), 3) on python 2, map returns a list, so the `all()` won't be able to short-circuit (it will need to go over all `args` first, whereas the genexp would stop at the first false)

dsl_parser/functions.py
```diff
@@ -806,10 +860,33 @@ def evaluate_outputs(outputs_def,
...
+        :param args: arguments that the function has been called with lastly.
+        :param kwargs: keyword arguments that the function has been called with
+            lastly.
+        :return: relevant string containing information about the arguments.
+        """
+        msg = "Limit was reached with the following path - {}"
+        if args and len(args) == 4:
+            return msg.format(args[3])
+        if kwargs and 'path' in kwargs:
+            msg.format(kwargs['path'])
```
tehasdf: return

dsl_parser/functions.py
```diff
@@ -850,63 +921,36 @@ def runtime_evaluation_handler(get_node_instances_method,
...
-        if isinstance(_func, Function):
-            _func.validate(plan)
-        if isinstance(_func, GetProperty):
-            get_property_functions.append(_func)
-            return _func
+        func = parse(v, scope=scope, context=context, path=path)
+        if isinstance(func, Function):
+            func.validate(plan)
+            level_of_tree_has_runtime_func.append(False)
+        scan.scan_properties(
```
tehasdf: this is essentially a depth-first search i think, but it's very complex, also due to how we keep a whole list of the `level_of_tree_has_runtime_func`? one bool should be enough i think, no?

i think it ends up being one bool anyway due to the `.pop` but we somewhat hide that; maybe we can replace this with just a `level[0] = level[0] or func.eval_type == RUNTIME` etc. (using a list with one element, which is a common idiom in python 2 with no `nonlocal`)
(or we could write a class and have the handler here be a method, and the bool be an instance attribute)

either way, this function could sure use some comments

dsl_parser/functions.py
```diff
@@ -954,3 +998,58 @@ def get_nested_attribute_value_of_capability(root, path):
...
+
+    def __exit__(self, tp, value, tb):
+        self.calls_cnt -= 1
+
+
+def limit_recursion(limit, args_to_str_func=None):
+    _recursion_limit_ctx = RecursionLimit(limit, args_to_str_func)
+
+    def decorator(f):
+        def wrapper(*args, **kwargs):
```
tehasdf: needs `@functools.wraps`

dsl_parser/tests/test_functions.py
```diff
@@ -1229,7 +1096,329 @@ def test_outputs(self):
...
+            {
+                'concat':
+                    ['one', 'value', {'get_attribute': ['node', 'attribute']}]
+            },
+            outputs['output3']['value'])
+
+
+class TestFunctionIntegration(AbstractTestParser):
+    """
+    VVV- REMOVE THIS -VVV
```
tehasdf: remove this! :P

dsl_parser/functions.py
davidginzbourg: Didn't know that (also forgot about generators)!

dsl_parser/functions.py
davidginzbourg: Oops. Done.

dsl_parser/functions.py
davidginzbourg: (Continued on Slack)

dsl_parser/functions.py
davidginzbourg: Done.

dsl_parser/tests/test_functions.py
davidginzbourg: Done.

PR #187, opened 2018-12-30
CY-821 restclient: executions.resume

tehasdf:


cloudify_rest_client/executions.py
```diff
@@ -255,3 +255,18 @@ def cancel(self, execution_id, force=False, kill=False):
...
+
+        :param execution_id: Id of the execution to resume.
+        :param force: Whether to resume failed/cancelled executions by
+                      retrying their failed tasks.
+        :return: Resumed execution.
+        """
+        uri = '/{self._uri_prefix}/{id}'.format(self=self, id=execution_id)
+        action = 'force-resume' if force else 'resume'
+        response = self.api.post(uri,
+                                 data={'action': action},
```
tyacbovi: Why pass a string and not a bool?

tehasdf: because executions' POST already works in terms of the 'action' command (it can take "cancel", "force-cancel", "kill", "resume", and now also "force-resume")

PR #188, opened 2018-12-30
CY-844 Mgmtworker: only allow resuming resumable operations

tehasdf:


tehasdf:
Currently contains #175

PR #199, opened 2019-01-13
Install/uninstall: skip NOP tasks and their events 

tehasdf:


cloudify/plugins/lifecycle.py
```diff
@@ -206,87 +225,128 @@ def install_node_instance_subgraph(instance, graph, **kwargs):
...
+            (configure or
+             [instance.send_event(
+                 'Configuring node instance: nothing to do')]) +
+            postconf +
+            (start or
+             [instance.send_event('Starting node instance: nothing to do')]) +
+            post_start +
+            monitoring_start +
+            establish +
+            started
```
tehasdf: dont use the "nothing to do" message here

PR #209, opened 2019-01-30
CY-1013 Expose TMPDIR to the script

tehasdf:
By default, expose the exec temp directory to the script, so that
`mktemp` calls inside the script use it


tehasdf:
port of cloudify-cosmo/cloudify-script-plugin#66

PR #214, opened 2019-02-09
CY-1028 : fixing the validation of remote resource scripts

tyacbovi:
* Remote resource (blueprint import) can have scripts, but in the current
  parsed script there is no need to verify those exists.
* Adding fix for namespced script plugin name in workflows
* Clearing up the error msg if there is no plugin or script resource in not
  found for an operation.

### DIFF
--- a/dsl_parser/elements/imports.py
+++ b/dsl_parser/elements/imports.py
@@ -181,12 +181,12 @@ def _extract_import_parts(import_url,
             namespaced, the returned namespace will be
             None.
     """
-    namespace_op_location = utils.find_namespace_location(import_url)
-    namespace = None
+    namespace, _, import_url = \
+        import_url.rpartition(constants.NAMESPACE_DELIMITER)
 
-    if namespace_op_location != -1:
-        namespace = import_url[:namespace_op_location]
-        import_url = import_url[namespace_op_location + 2:]
+    if namespace == '':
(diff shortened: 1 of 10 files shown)


dsl_parser/elements/imports.py
```diff
@@ -181,12 +181,10 @@ def _extract_import_parts(import_url,
...
     """
-    namespace_op_location = utils.find_namespace_location(import_url)
-    namespace = None
+    namespace, _, import_url = \
+        import_url.rpartition(constants.NAMESPACE_DELIMITER)
 
-    if namespace_op_location != -1:
-        namespace = import_url[:namespace_op_location]
-        import_url = import_url[namespace_op_location + 2:]
+    namespace = None if namespace == '' else namespace
```
barucoh: Anyway to make it better? i.e - not have "namespace = ... else namespace"

dsl_parser/utils.py
```diff
@@ -348,3 +348,7 @@ def is_blueprint_import(import_url):
 
 def remove_blueprint_import_prefix(import_url):
     return import_url.replace(constants.BLUEPRINT_IMPORT, '')
+
+
+def find_partial_matches_in_list(value, items):
```
barucoh: Please name it to be more user-friendly rather than 'value' and write a little description of it's usage

PR #216, opened 2019-02-19
CY-1056: Add test for node instance state

isaac-s:
This used to happen at the CLI side. As part of refactoring how filtering is done, I decided to move this check to the server side.


isaac-s:
@idob-cloudify right. I'll close this PR.

PR #218, opened 2019-02-20
CY-1071: deal with case when event/log message is not valid text

isaac-s:
A better approach would be to log the exception somewhere, but for that we need a Python logger configured, and we currently don't have it in that context.

### DIFF
--- a/cloudify/event.py
+++ b/cloudify/event.py
@@ -75,7 +75,12 @@ class Event(object):
 
     @property
     def text(self):
-        message = self._event['message']['text'].encode('utf-8')
+        message = self._event['message']['text']
+        try:
+            message = message.encode('utf-8')
+        except UnicodeDecodeError as ex:
+            message = "<Unable to encode as UTF-8; error message: {0}>".format(
+                ex)
         if self.is_log_message:
             message = '{0}: {1}'.format(self.log_level, message)
         elif (self.event_type in ('task_rescheduled', 'task_failed')):
--- a/cloudify/exceptions.py
+++ b/cloudify/exceptions.py
@@ -118,13 +118,17 @@ class CommandExecutionException(Exception):
         Exception.__init__(self, self.__str__())
 
     def __str__(self):
-        return "Command '{0}' executed with an error." \
-               "\ncode: {1}" \
-               "\nerror: {2}" \
-               "\noutput: {3}" \
-            .format(self.command, self.code,
-                    self.error or None,
-                    self.output or None)
+        return "Command '{0}' executed with error code {1}.\n" \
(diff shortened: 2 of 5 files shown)


cloudify/event.py
```diff
@@ -75,7 +75,12 @@ def operation_info(self):
 
     @property
     def text(self):
-        message = self._event['message']['text'].encode('utf-8')
+        message = self._event['message']['text']
+        try:
+            message = message.encode('utf-8')
+        except UnicodeDecodeError as ex:
+            message = "<Unable to encode to UTF-8; error mesage: {0}>".format(
+                str(ex))
```
tehasdf: no need for the `str()` call, using `.format` to format it into a string already does that

tehasdf: also this should be UnicodeEncodeError, which is what .encode throws, no?

isaac-s: @tehasdf I thought that too, but it actually raised `UnicodeDecodeError`

cloudify/exceptions.py
```diff
@@ -118,13 +118,17 @@ def __init__(self, command, error, output, code):
...
 
     def __str__(self):
-        return "Command '{0}' executed with an error." \
-               "\ncode: {1}" \
-               "\nerror: {2}" \
-               "\noutput: {3}" \
-            .format(self.command, self.code,
-                    self.error or None,
-                    self.output or None)
+        return "Command '{0}' executed with error code {1}.\n" \
```
isaac-s: Just added this on the way.... better formatting of this error message

cloudify/utils.py
```diff
@@ -374,8 +374,12 @@ def run(self, command,
 
         :param command: The command to execute.
         :param exit_on_failure: False to ignore failures.
-        :param stdout_pipe: False to not pipe the standard output.
-        :param stderr_pipe: False to not pipe the standard error.
+        :param stdout_pipe: False to not pipe the standard output. In this
```
isaac-s: Improved comment (also unrelated to the original change)

cloudify/event.py
```diff
@@ -75,7 +75,12 @@ def operation_info(self):
 
     @property
     def text(self):
-        message = self._event['message']['text'].encode('utf-8')
+        message = self._event['message']['text']
+        try:
+            message = message.encode('utf-8')
```
tehasdf: encoding here is still wrong

isaac-s: @tehasdf as discussed, while wrong, this should be resolved by a different JIRA (or in the Py3 branch).

geokala:

PR #222, opened 2019-02-27
Work with external rabbitmq

geokala:


cloudify/cluster.py
```diff
@@ -113,12 +113,13 @@ def set_cluster_active(node, filename=None):
     set_cluster_settings(settings, filename=filename)
 
 
-def get_cluster_amqp_settings():
-    active = get_cluster_active()
-    if active:
```
geokala: We no longer need to check if they're active, as we're requiring rabbit to be external and clustered.

PR #223, opened 2019-02-27
CY-1084 Add the “Execution-Token” header to the rest client

inbala:
* Create a new header “Execution-Token” when creating the rest client during a execution

* Add get_execution_token_from_request to the utils


cloudify_rest_client/client.py
```diff
@@ -106,8 +108,12 @@ def has_kerberos(self):
         return bool(HTTPKerberosAuth) and is_kerberos_env()
 
     def has_auth_header(self):
-        return CLOUDIFY_TOKEN_AUTHENTICATION_HEADER in self.headers or \
-               CLOUDIFY_AUTHENTICATION_HEADER in self.headers
+        auth_headers = [CLOUDIFY_TOKEN_AUTHENTICATION_HEADER,
+                        CLOUDIFY_AUTHENTICATION_HEADER,
+                        CLOUDIFY_API_AUTH_TOKEN_HEADER,
+                        CLOUDIFY_EXECUTION_TOKEN_HEADER]
+        return any([header in self.headers
```
tehasdf: can skip the `[]`, making this just `return any(header in self.headers for header in auth_headers)`, so that the whole list won't need to be precomputed

cloudify/utils.py
```diff
@@ -242,7 +242,9 @@ def get_execution_token():
     Returns the execution token to use when calling the REST service
     """
     try:
-        return _get_current_context().execution_token
+        context = _get_current_context()
+        return context.execution_token if hasattr(context, 'execution_token') \
```
tehasdf: hmm i dont understand, why would there be no attribute? there is a property now, so the hasattr should have no reason to be here?

tehasdf: is it for WorkflowContext? maybe we also need to add the property there?

inbala: Ah yes I added it because it was failing when the context was `CancelCloudifyContext`, I thought I shouldn't added it there, but maybe I do?

PR #227, opened 2019-03-04
Add managers table rest client

barucoh:


cloudify_rest_client/manager.py
```diff
@@ -114,6 +161,41 @@ def put_config(self, name, value, force=False):
...
         return ConfigItem(response)
 
+    def get_manager(self, hostname):
+        """
+        Get a specific manager from the managers table
+        :param hostname: manager to get
+        """
+        if not hostname:
+            raise ValueError('Enter a valid hostname')
+        response = self.api.get('/managers/{0}'.format(hostname))
```
tyacbovi: I would join this line with the next line

tyacbovi:
Also do not forget to update rest docs in the future

cloudify_rest_client/manager.py
barucoh: I think following the standard with the rest of the file will keep it readable.

PR #229, opened 2019-03-04
Moving the managers client to work with 1 endpoint and adding remove_…

barucoh:
…manager client

### DIFF
--- a/cloudify_rest_client/manager.py
+++ b/cloudify_rest_client/manager.py
@@ -185,16 +185,6 @@ class ManagerClient(object):
         })
         return ConfigItem(response)
 
-    def get_manager(self, hostname):
-        """
-        Get a specific manager from the managers table
-        :param hostname: manager to get
-        """
-        if not hostname:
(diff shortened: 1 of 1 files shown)


cloudify_rest_client/manager.py
```diff
@@ -212,12 +202,30 @@ def add_manager(self, hostname, private_ip, public_ip, version,
...
+        the cluster, not necessarily for uninstalling the manager
+        :param hostname: The manager's hostname
         """
-        Get all the managers in the managers table
+        response = self.api.delete('/managers', data={
+            'hostname': hostname
+        })
+        return ManagerItem(response)
+
+    def get_managers(self, hostname=None, _include=None):
```
tyacbovi: What can _include will contain here other them hostnames?
Can you use _include to pass the specific hostname to the query only?

barucoh: Include only specific columns in the returned value

PR #230, opened 2019-03-05
Adding update method to managers client and correcting according to u…

barucoh:
…pdated DB model

### DIFF
--- a/cloudify_rest_client/manager.py
+++ b/cloudify_rest_client/manager.py
@@ -77,10 +77,18 @@ class ManagerItem(dict):
         self.update(manager)
 
     @property
+    def id(self):
+        """
+        Manager's ID in the DB (unique, incremental)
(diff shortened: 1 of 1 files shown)


cloudify_rest_client/manager.py
```diff
@@ -76,6 +76,14 @@ def __init__(self, manager):
         super(ManagerItem, self).__init__()
         self.update(manager)
 
+    @property
+    def id(self):
+        """
+        Manager's ID in the DB (unique, incremental)
+        type: db.Integer
```
tehasdf: let's change the type in the docstrings for this whole class to not use db.Text/db.Integer, those types are restservice-specific and are meaningless for the client (use just int/str instead)

barucoh: Fixed

PR #235, opened 2019-03-13
Add 'expired' field to Cloudfiy license

adigrabow:


cloudify_rest_client/license.py
```diff
@@ -73,6 +73,13 @@ def signature(self):
         """
         return self.get('signature')
 
+    @property
+    def expired(self):
+        """
+        :returns:Whether or not this Cloudify license has expired
```
inbala: A white space is missing :P

PR #259, opened 2019-04-07
CY-933 Fix input validations

davidginzbourg:
- Fixes regex string validation: it now uses basestring and not str.
- Fixes basic data types not validated during deployment creation.

### DIFF
--- a/dsl_parser/tests/test_inputs.py
+++ b/dsl_parser/tests/test_inputs.py
@@ -840,8 +840,10 @@ inputs:
         self.assertEqual(1, len(parsed[consts.INPUTS]))
         self.assertEqual('hi',
                          parsed[consts.INPUTS]['some_input'][consts.DEFAULT])
-        plan = prepare_deployment_plan(parsed,
-                                       inputs={'some_input': 'ab'})
+        plan = prepare_deployment_plan(parsed, inputs={'some_input': 'ab'})
+        self.assertEqual(1, len(plan[consts.INPUTS]))
+        self.assertEqual('ab', plan[consts.INPUTS]['some_input'])
+        plan = prepare_deployment_plan(parsed, inputs={'some_input': u'ab'})
         self.assertEqual(1, len(plan[consts.INPUTS]))
         self.assertEqual('ab', plan[consts.INPUTS]['some_input'])
 
@@ -1140,7 +1142,7 @@ data_types:
             ERROR_MISSING_PROPERTY)
         self.assertIn('is missing property b', ex.message)
 
-    def test_input_validate_type_raises(self):
+    def test_input_validate_non_existing_type(self):
         yaml = self.BASIC_VERSION_SECTION_DSL_1_2 + """
 inputs:
     some_input:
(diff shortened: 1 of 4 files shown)


dsl_parser/tests/test_inputs.py
```diff
@@ -1152,80 +1154,121 @@ def test_input_validate_type_raises(self):
...
+
+    def test_input_validate_boolean_raises_with_deployment(self):
+        self._test_input_validate_type_raises_with_deployment('boolean', 'x')
+
+    def test_input_validate_string_successful(self):
+        self._test_input_validate_type_successful('string', 'some_string',
+                                                  self.assertEqual)
+        # Using type: string, allows us to pass any kind of type (due to
+        # backward compatibility)
+        self._test_input_validate_type_successful('string', list(),
```
tyacbovi: why does "list()" works as a type name?

dsl_parser/tests/test_inputs.py
```diff
@@ -1152,80 +1154,121 @@ def test_input_validate_type_raises(self):
...
+        self._test_input_validate_type_raises_with_deployment('float', True)
+
+    def test_input_validate_boolean_successful(self):
+        self._test_input_validate_type_successful('boolean', False,
+                                                  self.assertEqual)
+
+    def test_input_validate_boolean_raises(self):
+        self._test_input_validate_type_raises('boolean', [])
+
+    def test_input_validate_boolean_raises_with_deployment(self):
```
tyacbovi: improve tests names, they are not reflecting the cause that the test failed

dsl_parser/tests/test_inputs.py
```diff
@@ -1152,80 +1154,121 @@ def test_input_validate_type_raises(self):
...
+    def test_input_validate_integer_raises_with_deployment(self):
+        self._test_input_validate_type_raises_with_deployment('integer', 'x')
+        self._test_input_validate_type_raises_with_deployment('integer', True)
+
+    def test_input_validate_float_successful(self):
+        self._test_input_validate_type_successful('float', 123,
+                                                  self.assertEqual)
+        self._test_input_validate_type_successful('float', 123.1,
+                                                  self.assertEqual)
+        self._test_input_validate_type_successful('float', long(),
```
tyacbovi: the other way will work? because u lose accuracy

dsl_parser/tests/test_inputs.py
```diff
@@ -1152,80 +1154,121 @@ def test_input_validate_type_raises(self):
...
+        self._test_input_validate_type_raises('boolean', [])
+
+    def test_input_validate_boolean_raises_with_deployment(self):
+        self._test_input_validate_type_raises_with_deployment('boolean', 'x')
+
+    def test_input_validate_string_successful(self):
+        self._test_input_validate_type_successful('string', 'some_string',
+                                                  self.assertEqual)
+        # Using type: string, allows us to pass any kind of type (due to
+        # backward compatibility)
```
tyacbovi: backward compatibility of a Cloudify feature?

dsl_parser/tests/test_inputs.py
davidginzbourg: It's the the type name, it's the value, i.e. an empty list.

PR #270, opened 2019-04-25
CY-1207 Add sites client

inbala:


cloudify_rest_client/sites.py
```diff
@@ -0,0 +1,167 @@
+########
+# Copyright (c) 2013-2019 GigaSpaces Technologies Ltd. All rights reserved
```
tyacbovi: Cloudify

cloudify_rest_client/sites.py
```diff
@@ -0,0 +1,167 @@
...
+
+        :param _include: List of fields to include in response.
+        :param sort: Key for sorting the list.
+        :param is_descending: True for descending order, False for ascending.
+        :param kwargs: Optional filter fields. For a list of available fields
+               see the REST service's models.Site.fields
+        :return: Sites list.
+        """
+
+        params = kwargs
```
tyacbovi: I think it's less important to rename this

cloudify_rest_client/sites.py
inbala: Oops

PR #273, opened 2019-04-29
Dump/restore for local tasks 

tehasdf:


cloudify/workflows/workflow_context.py
```diff
@@ -1234,6 +1216,27 @@ def store_operation(self, graph_id, dependencies,
     def remove_operation(self, operation_id):
         raise NotImplementedError('Implemented by subclasses')
 
+    # factory methods for local tasks
+    # TODO: if possible, consider inlining those to simplify
```
geokala: Is this something that we are going TODO or should the comment be splatted?

PR #274, opened 2019-04-29
Support multiple REST backends

tehasdf:


cloudify/cluster.py
```diff
@@ -13,158 +13,26 @@
...
-
-
 class ClusterHTTPClient(HTTPClient):
     default_timeout_sec = 5
     retries = 30
     retry_interval = 3
 
     def __init__(self, *args, **kwargs):
         super(ClusterHTTPClient, self).__init__(*args, **kwargs)
+        hosts = self.host if isinstance(self.host, list) else [self.host]
```
tehasdf: maybe add a comment

cloudify/cluster.py
```diff
@@ -13,158 +13,26 @@
...
+        self.hosts = itertools.cycle(hosts)
```
geokala: Probably good to randomise the order here so that we don't end up with one manager taking every single request until it goes down (a rudimentary sort of load balancing).

cloudify/cluster.py
geokala: OK, adding a comment: I'm not convinced a comment in code is required, since this line is pretty self explanatory?

PR #288, opened 2019-05-08
CY-1253 Add api for setting site for a deployment

inbala:


cloudify_rest_client/deployments.py
```diff
@@ -296,3 +314,21 @@ def set_visibility(self, deployment_id, visibility):
...
+        Updates the deployment's site
+
+        :param deployment_id: Deployment's id to update.
+        :param site_name: The site to update
+        :param detach_site: True for detaching the current site, making the
+                            deployment siteless
+        :return: The deployment.
+        """
+        data = {'detach_site': detach_site}
+        if site_name:
```
tyacbovi: What will happen if detach-site and site are equal?
maybe in that case a warning message will be printed and save the http call?

PR #290, opened 2019-05-10
Scale up: refresh node instances

tehasdf:


tehasdf:
hopefully at some point we can get to refactoring the `WorkflowNodesAndInstancesContainer` altogether, ideally to not use multiple inheritance, which will make this whole business way cleaner

PR #293, opened 2019-05-12
CY-1219-import-export-secrets

YoniItzhak:


cloudify/cryptography_utils.py
```diff
@@ -53,6 +56,22 @@ def _get_encryption_key():
         return rest_security_conf['encryption_key']
 
 
+def generate_key_using_password(password):
```
adigrabow: I think `salt` should be a param here (with None as default)

cloudify/cryptography_utils.py
```diff
@@ -53,6 +56,22 @@ def _get_encryption_key():
...
+    password = password.encode()
+    salt = b'salt_'
+    # should verify it is safe enough, option: generate a random one and
```
adigrabow: Cab delete this comment

cloudify_rest_client/secrets.py
```diff
@@ -126,6 +133,17 @@ def get(self, key):
...
 
+    def export(self, **kwargs):
+        """
+        Returns a list of secrets to be exported
+
+        :param kwargs: the parameters given from the user
+        :return: Secrets' list
+        """
+        params = kwargs
+        response = self.api.get('/secrets/command/export', params=params)
```
adigrabow: Just change the endpoint name

cloudify_rest_client/secrets.py
```diff
@@ -126,6 +133,17 @@ def get(self, key):
         response = self.api.get('/secrets/{0}'.format(key))
         return Secret(response)
 
+    def export(self, **kwargs):
+        """
+        Returns a list of secrets to be exported
```
adigrabow: Do you think it's clear enough? Why not `exports secrets to file`?

YoniItzhak: I think it's a good idea, but won't it be a bit confusing when all the other functions are named the same as in the cli file?

PR #314, opened 2019-06-02
New command to remove the 'installing' prefix from

adigrabow:
plugins archive_name


cloudify_rest_client/plugins.py
```diff
@@ -294,3 +294,13 @@ def set_visibility(self, plugin_id, visibility):
             '/plugins/{0}/set-visibility'.format(plugin_id),
             data=data
         )
+
+    def finish_installation(self, plugin_id):
+        """
+        Removes the `installing-` prefix from the `archive_name` of a plugin.
```
tyacbovi: Also here I will not mention the impl detail in the doc

PR #320, opened 2019-06-13
RequestResponseHandlers: make sure amqp queues are deleted

tehasdf:


tehasdf:
This is some overall simplifying of the request/response handlers, and also making sure that an issue with a queue leak doesn't exist

tehasdf:
the accompanying cloudify-agent PR is cloudify-cosmo/cloudify-agent#554

PR #321, opened 2019-06-13
CY-1060 Add the plugins update API to client

davidginzbourg:
- Adds an update_plugins flag to the deployment update API calls

### DIFF
--- a/cloudify/utils.py
+++ b/cloudify/utils.py
@@ -16,6 +16,7 @@
 import os
 import ssl
 import sys
+import time
 import pika
 import shlex
 import random
@@ -675,19 +676,22 @@ def extract_plugins_to_install(plugin_list, filter_func):
     return [p for p in plugin_list if p[PLUGIN_INSTALL_KEY] and filter_func(p)]
 
 
-def extract_and_merge_plugins_to_install(deployment_plugins,
-                                         workflow_plugins,
-                                         filter_func=lambda _: True):
-    """Extracts plugins to install and merges them.
-
+def extract_and_merge_plugins(deployment_plugins,
+                              workflow_plugins,
+                              filter_func=lambda _: True,
+                              with_repetition=False):
+    """
     :param deployment_plugins: deployment plugins to install.
     :param workflow_plugins: workflow plugins to install.
     :param filter_func: predicate function to filter the plugins, should return
      True for plugins that don't need to be filtered out.
-    :return: the merged plugins to install.
+    :param with_repetition: can plugins appear twice in the list.
+    :return: a list of merged plugins that were marked for installation.
     """
     dep_plugins = extract_plugins_to_install(deployment_plugins, filter_func)
     wf_plugins = extract_plugins_to_install(workflow_plugins, filter_func)
+    if with_repetition:
+        return dep_plugins + wf_plugins
     return merge_plugins(dep_plugins, wf_plugins)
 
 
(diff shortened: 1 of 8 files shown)


cloudify/utils.py
```diff
@@ -753,3 +754,20 @@ def add_plugins_to_uninstall(ctx, plugins_to_uninstall, sequence):
             kwargs={
                 'plugins': plugins_to_uninstall,
                 'delete_managed_plugins': False}))
+
+
+def wait_for(callable_obj,
```
tyacbovi: why not to put this in the manger only? and maybe this already exists because this is very common

cloudify_rest_client/plugins_update.py
```diff
@@ -0,0 +1,132 @@
...
+#  * limitations under the License.
+
+from cloudify_rest_client.responses import ListResponse
+
+
+class PluginsUpdate(dict):
+
+    def __init__(self, update):
+        self.update(update)
+
```
tyacbovi: force is missing

cloudify_rest_client/plugins_update.py
```diff
@@ -0,0 +1,132 @@
...
+        response = self.api.post(
+            '/{self._uri_prefix}/{}/update/initiate'.format(blueprint_id,
+                                                            self=self),
+            params=params
+        )
+        return PluginsUpdate(response)
+
+    def finalize_plugins_update(self, plugins_update_id):
+        """
+        Finalize the plugins update.
```
tyacbovi: is internal or for client use?

cloudify/utils.py
davidginzbourg: I just moved it from a different place and used it. The other functions are in tests and cli, all separated. I think it's out of scope to clean it up and centralize it.

cloudify_rest_client/plugins_update.py
davidginzbourg: Force is not saved under a plugin update, it's just to tell the manager to ignore other running updates.

cloudify_rest_client/plugins_update.py
davidginzbourg: Internal but used by the plugins update workflow

cloudify/utils.py
```diff
@@ -675,19 +676,22 @@ def extract_plugins_to_install(plugin_list, filter_func):
     return [p for p in plugin_list if p[PLUGIN_INSTALL_KEY] and filter_func(p)]
 
 
-def extract_and_merge_plugins_to_install(deployment_plugins,
-                                         workflow_plugins,
-                                         filter_func=lambda _: True):
-    """Extracts plugins to install and merges them.
-
+def extract_and_merge_plugins(deployment_plugins,
```
davidginzbourg: Please note I changed the func name here

cloudify_rest_client/plugins_update.py
tyacbovi: I think we need to because the use while looking at the history will want to know this

cloudify_rest_client/plugins_update.py
tyacbovi: also add here a note about this being internal and in the rest-docs and cli docs

cloudify_rest_client/plugins_update.py
davidginzbourg: I added it

cloudify_rest_client/plugins_update.py
davidginzbourg: There's no CLI command for this (neither did deployments update). Updated the docstring

PR #322, opened 2019-06-16
RequestResponse: response queue names start with response_

tehasdf:
This is because other amqp users are only allowed to access queues
named like that, note in https://github.com/cloudify-cosmo/cloudify-manager/blob/master/rest-service/manager_rest/amqp_manager.py#L198


tehasdf:
sanity CI is https://circleci.com/workflow-run/3720aeaf-d1c2-42fa-b0e5-081eb78c262a

PR #336, opened 2019-07-09
CY-1043 Get rabbitmq credentials from rest when sending tasks

tehasdf:
Instead of storing the credentials in operation context, which
might be stored in places that we don't want to store credentials in
(eg. the operations table), get the credentials via rest when
sending.

Also declare the credential fields on the restclient Tenant object -
they were there before (since CY-599), but undeclared

### DIFF
--- a/cloudify/utils.py
+++ b/cloudify/utils.py
@@ -266,10 +266,13 @@ def get_execution_token():
 
 
 def get_tenant():
-    """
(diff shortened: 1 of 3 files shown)


tehasdf:
TODO this requires additional mocks so that cloudify-agent's CI may pass

PR #338, opened 2019-07-11
Cy 1447 improve agents list

glukhman:


cloudify_rest_client/agents.py
```diff
@@ -142,15 +142,17 @@ def __init__(self, api):
         self._wrapper_cls = Agent
 
     def list(self, deployment_id=None, node_ids=None, node_instance_ids=None,
-             install_methods=None):
+             install_methods=None, all_tenants=False):
```
inbala: For consistency use kwargs with `_all_tenants` like other resources list

PR #339, opened 2019-07-12
Add cause to determine workflow failing event

kfrukacz-gs:
Extra cause will be appended to the causes list
for the failed task in case of workflow terminating failure.
The type will always be NonRecoverable error (regardles of type
of the original exception)


kfrukacz-gs:
FR to be created for this to handle it inside the model

kfrukacz-gs:
https://cloudifysource.atlassian.net/browse/CYBL-1032

PR #341, opened 2019-07-22
CY-1525 execute_operation: add operation name to the graph name

tehasdf:
This means that usages of execute_operation (in the restart workflow)
will have distinct names by default, because they run different
operations.

This doesn't change anything for other usages of execute_workflow,
because the name is just an opaque identifier, nothing really cares
if it's "execute_operation" or "execute_operation_cloudify.lifecycle.create"

Additionally, that means that looking at the operations table in the
db will be nicer, with this info in the name

### DIFF
--- a/cloudify/plugins/workflows.py
+++ b/cloudify/plugins/workflows.py
@@ -643,11 +643,11 @@ def _make_execute_operation_graph(ctx, operation, operation_kwargs,
 
 
 @workflow(resumable=True)
-def execute_operation(ctx, *args, **kwargs):
+def execute_operation(ctx, operation, *args, **kwargs):
     """ A generic workflow for executing arbitrary operations on nodes """
-
+    name = 'execute_operation_{0}'.format(operation)
     graph = _make_execute_operation_graph(
-        ctx, name='execute_operation', *args, **kwargs)
+        ctx, operation, name=name, *args, **kwargs)
     graph.execute()
 
 
--- a/cloudify/workflows/tasks_graph.py
+++ b/cloudify/workflows/tasks_graph.py
@@ -23,6 +23,7 @@ import networkx as nx
 from cloudify.workflows import api
 from cloudify.workflows import tasks
 from cloudify.state import workflow_ctx
+from cloudify.exceptions import NonRecoverableError
 
 
 def make_or_get_graph(f):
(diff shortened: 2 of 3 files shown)


cloudify/plugins/workflows.py
```diff
@@ -643,11 +643,11 @@ def _make_execute_operation_graph(ctx, operation, operation_kwargs,
 
 
 @workflow(resumable=True)
-def execute_operation(ctx, *args, **kwargs):
+def execute_operation(ctx, operation, *args, **kwargs):
     """ A generic workflow for executing arbitrary operations on nodes """
-
+    name = 'execute_operation_{0}'.format(operation)
```
tyacbovi: Can we add tests also for this?

tehasdf: no, because operations are only stored when using a manager; when using local, they're not stored, and so this name is passed nowhere

that's why i added integration tests - because i couldn't test this on the unittest side

cloudify/workflows/tasks_graph.py
```diff
@@ -68,7 +68,11 @@ def restore(cls, workflow_context, retrieved_graph):
             if op.containing_subgraph:
                 subgraph_id = op.containing_subgraph
                 op.containing_subgraph = None
-                subgraph = ops[subgraph_id]
+                # subgraph can be missing if the operation was failed but
+                # the subgraph was suceeded (which happens only if an
+                # operation runs twice at the same time, which shouldn't
+                # happen, but let's not break in that case)
+                subgraph = ops.get(subgraph_id, graph)
```
tehasdf: added this

this only really happens if we try to cancel+resume an operation that didn't get properly cancelled. Then we end up with 2 instances of the same operation running at the same time (which is pretty much an error condition and perhaps we should add a validation that this doesnt happen), and when the first operation finishes and marks the subgraph as done, and we try to resume the 2nd instance of the operation, the subgraph is already done. So, we'll just attach that operation to the main graph, in place of the subgraph.

tyacbovi: What do u think about adding some debug msg about this for future problems?

cloudify/workflows/tasks_graph.py
```diff
@@ -68,7 +68,11 @@ def restore(cls, workflow_context, retrieved_graph):
             if op.containing_subgraph:
                 subgraph_id = op.containing_subgraph
                 op.containing_subgraph = None
-                subgraph = ops[subgraph_id]
+                # subgraph can be missing if the operation was failed but
```
tyacbovi: The explanation is not as clear as you wrote in ur comment

PR #356, opened 2019-08-21
CY-1507 Allow get_input and get_property at runtime

tehasdf:


dsl_parser/functions.py
```diff
@@ -892,30 +860,98 @@ def handler(v, scope, context, path):
...
+            raise KeyError('Node {0} does not exist'.format(node_name))
+
+
+class _RuntimeEvaluationHandler(_EvaluationHandler):
+    def __init__(self, storage):
+        self._storage = storage
+        self._nodes_cache = {}
+        self._node_to_node_instances = {}
+        self._node_instances_cache = {}
+        self._secrets_cache = {}
```
tyacbovi: I am not about caching entities that could change during the lifespan of a deployment like capabilities, node instance (in scaling policy, dep update will de-validate this?, secrets too)

cloudify/workflows/local.py
```diff
@@ -416,22 +416,31 @@ def update_node_instance(self,
                 instance['state'] = state
             self._store_instance(instance)
 
-    def _get_node_instance(self, node_instance_id):
+    def _get_node_instance(self, node_instance_id, evaluate_functions=True):
         instance = self._load_instance(node_instance_id)
         if instance is None:
             raise RuntimeError('Instance {0} does not exist'
```
tyacbovi: Let's change also this to KeyError to keep the consistency

cloudify_rest_client/deployments.py
```diff
@@ -260,6 +261,10 @@ def create(self,
                                 plugins containing source URL will
                                 be installed from source.
         :param site_name: The name of the site for the deployment.
+        :param functions_eager: Whether the get_property and get_input
```
tyacbovi: well, I think we don't want to specifically address get_input and get_property, but address them as static/non-runtime/deployment-creation-elevated functions because we could add more of this kind of functions in the future. And I am not sure about the flag, I think the special behavior is the runtime eval so I would think that will be something to turn on like what a user will need to do for this.

dsl_parser/tasks.py
```diff
@@ -144,12 +144,13 @@ def _validate_secrets(plan, get_secret_method):
...
-def prepare_deployment_plan(plan, get_secret_method=None, inputs=None, **_):
+def prepare_deployment_plan(plan, get_secret_method=None, inputs=None,
+                            functions_eager=True, **_):
     """
     Prepare a plan for deployment
     """
     plan = models.Plan(copy.deepcopy(plan))
     _set_plan_inputs(plan, inputs)
-    _process_functions(plan)
+    _process_functions(plan, functions_eager)
```
tyacbovi: if we will stick with this flag, I think maybe to switch this to eager_evaluation

dsl_parser/functions.py
```diff
@@ -303,9 +292,10 @@ def validate(self, plan):
         args = [self.node_name] + self.property_path
         if any(is_function(x) for x in args):
             return
-        self.evaluate(plan)
+        handler = _PlanEvaluationHandler(plan, eager=True)
```
tyacbovi: what do u think about passing evaluate func handler and plan, so we would not re create _PlanEvaluationHandler in every call of validate func?

dsl_parser/tests/test_get_attribute.py
```diff
@@ -354,48 +344,44 @@ def _test_process_attribute_scaling_group_ambiguity_resolution(
                     for r in node_instance.get('relationships', [])],
             })
 
-        def get_node_instances(node_id):
-            return node_to_node_instances[node_id]
+        class Storage(object):
```
tyacbovi: I think u can generalize the storage so it can be more easy to write test with this

dsl_parser/tests/test_functions.py
```diff
@@ -1251,134 +1250,6 @@ def test_func_integrations_within_list(self):
...
-                            - get_attribute: [ some_node, dummy_prop ]
-                            - get_input: dummy_input
-    """
-        with self.assertRaisesRegexp(
-                exceptions.FunctionValidationError,
-                'Runtime function .+ cannot be nested within a non-runtime '
-                'function \\(found in .+\\)'):
-            prepare_deployment_plan(self.parse_1_1(yaml))
-
-    def test_get_property_doesnt_accept_runtime_func_as_args(self):
```
tyacbovi: so what about tests that test this works :)

dsl_parser/tests/test_get_attribute.py
```diff
@@ -98,30 +98,31 @@ class TestEvaluateFunctions(AbstractTestParser):
...
+                return [self.get_node_instance(node_id)]
+
+            def get_node_instance(self, node_instance_id):
+                result = NodeInstance({
+                    'id': node_instance_id,
+                    'node_id': 'webserver',
+                    'runtime_properties': {
+                    }
+                })
+                if node_instance_id == 'node1':
```
tyacbovi: what creating a dict for this

dsl_parser/tests/test_register_function.py
```diff
@@ -80,29 +77,29 @@ def validate(self, plan):
...
+                return self.get_node_instances()[0]
+
+            def get_node(self, node_id):
+                return Node({'id': node_id})
+
+        storage = Storage()
+        storage.get_secret = self._get_secret_mock
+        storage.get_capability = self._get_capability_mock
+
+        o = functions.evaluate_outputs(parsed['outputs'], storage)
```
tyacbovi: let's change 'o' to something more meaningful

dsl_parser/tests/test_functions.py
```diff
@@ -718,6 +718,57 @@ def get_property_from_get_property(self):
...
+        properties:
+            a: {get_property: [vm2, c]}
+            c: secret1
+    vm2:
+        type: vm_type
+        properties:
+            a: xxx
+            c: secret2
+"""
+        plan = prepare_deployment_plan(self.parse(yaml), functions_eager=False)
```
tyacbovi: and eager tests that fail with runtime ?

dsl_parser/functions.py
```diff
@@ -308,9 +308,13 @@ def get_node_template(self, handler):
                     '{0} can only be used within a relationship but is used '
                     'in {1}'.format(self.node_name, self.path))
             if self.node_name == SOURCE:
-                node = self.context['node_template']
+                # this is messy, see CY-1676 to fix it
+                node = self.context.get('node_template') or \
```
tyacbovi: maybe put some doc also here, so it will be more clear without the Jira

cloudify/workflows/local.py
```diff
@@ -424,22 +416,31 @@ def update_node_instance(self,
...
-                               .format(node_id))
+            raise KeyError('Node {0} does not exist'.format(node_id))
+        if evaluate_functions:
+            dsl_functions.evaluate_node_functions(node, self)
         return copy.deepcopy(node)
 
-    def get_nodes(self):
-        return copy.deepcopy(self._nodes.values())
+    def get_nodes(self, evaluate_functions=True):
+        nodes = copy.deepcopy(self._nodes.values())
```
geokala: Maybe only do the deepcopy after deciding whether we can short circuit due to evaluate_functions=False?

cloudify/workflows/local.py
```diff
@@ -545,10 +550,14 @@ def _load_instance(self, node_instance_id):
...
-    def get_node_instances(self, node_id=None):
+    def get_node_instances(self, node_id=None, evaluate_functions=True):
         instances = self._node_instances.values()
         if node_id:
             instances = [i for i in instances if i.node_id == node_id]
+        if evaluate_functions:
+            for instance in instances:
+                dsl_functions.evaluate_node_instance_functions(
+                    instance, self)
         return copy.deepcopy(instances)
```
geokala: Any idea why this returns a deepcopy? It seems like we modify stuff in place and then return a copy for some reason?

cloudify/workflows/local.py
```diff
@@ -675,11 +686,15 @@ def _store_instance(self, node_instance, lock=True):
...
-    def get_node_instances(self, node_id=None):
+    def get_node_instances(self, node_id=None, evaluate_functions=True):
         instances = [self._get_node_instance(instance_id)
                      for instance_id in self._instance_ids()]
         if node_id:
             instances = [i for i in instances if i.node_id == node_id]
+        if evaluate_functions:
+            for instance in instances:
+                dsl_functions.evaluate_node_instance_functions(
+                    instance, self)
```
geokala: There seems to be a lot of code duplication between FileStorage and InMemoryStorage. Is this something that can reasonably be refactored or would that be massively more hassle than it's worth at the moment?

cloudify_rest_client/deployments.py
geokala: It is using the behaviour you described when this param is at its default setting already- you have to turn off functions_eager in order to get the new (runtime evaluation) behavious.

cloudify/workflows/local.py
geokala: Just discussed- this result is consumed by plugins, etc, so we need to deepcopy to avoid unexpected mutability issues.

dsl_parser/functions.py
```diff
@@ -35,6 +35,63 @@
...
+class RecursionLimit(object):
+    def __init__(self, limit, args_to_str_func=None):
+        """Initializes the recursion limit context manager. This context
+        manager limits the recursion only for this specific function, unlike
+        the sys.setrecursionlimit(limit) function, which sets the call stack
+        depth.
+
+        :param limit: recursion limit (inclusive).
+        :param args_to_str_func: function that returns a string containing
+            information about the last arguments used. It's signature should
```
geokala: <pedant> Its signature should be... 
Possessive it does not require apostrophes.
</pedant>

dsl_parser/functions.py
```diff
@@ -228,14 +226,11 @@ def validate(self, plan):
                 "get_input function references an "
                 "unknown input '{0}'.".format(input_value))
 
-    def evaluate(self, plan):
+    def evaluate(self, handler):
         if isinstance(self.input_value, list):
-            return self._get_input_attribute(plan.inputs[self.input_value[0]])
-        if self.input_value not in plan.inputs:
-            raise exceptions.UnknownInputError(
```
geokala: Presumably this failure condition is now handled in the handler? (I'm asking this in case I forget to verify that once I hit the relevant section)

dsl_parser/functions.py
```diff
@@ -772,115 +734,94 @@ def parse(raw_function, scope=None, context=None, path=None):
...
     """
-    handler = runtime_evaluation_handler(get_node_instances_method,
-                                         get_node_instance_method,
-                                         get_node_method,
-                                         get_secret_method,
-                                         get_capability_method)
+    scope = None
+    if 'source' in context and 'target' in context:
+        scope = scan.NODE_TEMPLATE_RELATIONSHIP_SCOPE
+    elif 'self' in context:
```
geokala: Can this ever fall through?

dsl_parser/functions.py
```diff
@@ -772,115 +734,94 @@ def parse(raw_function, scope=None, context=None, path=None):
...
-        get_node_method=get_node_method,
-        get_secret_method=get_secret_method,
-        get_capability_method=get_capability_method)
+        storage=storage)
 
 
-def _handler(evaluator, **evaluator_kwargs):
-    def _args_to_str_func(args, kwargs):
-        """Used to display extra information when recursion limit is reached.
+def _args_to_str_func(handler, v, scope, context, path):
```
geokala: This doesn't match the signature you said in the RecursionLimit class where you announced it should be fun(args, kwargs)

dsl_parser/functions.py
geokala: It removed by pedant tags!

dsl_parser/rel_graph.py
```diff
@@ -844,9 +855,11 @@ class Context(object):
     def __init__(self,
                  plan_node_graph,
                  deployment_node_graph,
+                 plan,
                  previous_deployment_node_graph=None,
                  previous_deployment_contained_graph=None,
                  modified_nodes=None):
+        self.plan = plan
```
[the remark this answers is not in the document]
tehasdf: object size doesn't matter in python for setting an attribute

but this was removed anyway after i realized that node instance count changes don't need to be part of this pr

dsl_parser/functions.py
```diff
@@ -803,84 +796,59 @@ def evaluate_functions(payload, context,
...
+def _args_to_str_func(args, kwargs):
+    """Used to display extra information when recursion limit is reached.
+
+    :param args: arguments that the function has been called with lastly.
+    :param kwargs: keyword arguments that the function has been called with
+        lastly.
+    :return: relevant string containing information about the arguments.
+    """
+    msg = "Limit was reached with the following path - {0}"
+    if args and len(args) == 5:
```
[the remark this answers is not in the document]
tehasdf: removed them entirely, the function is one-off by design

dsl_parser/functions.py
tehasdf: this cache lives for the duration of a single evaluate call, not for the duration of the whole deployment

and it worked before - this cache isn't new..

cloudify_rest_client/deployments.py
tehasdf: renamed to `runtime_only_evaluation`

dsl_parser/tests/test_functions.py
tehasdf: re-added that validation and added a test

old test did `prepare_deployment_plan(parse())` but it actually threw the error inside of parse, so the prepare was useless

however now the error is thrown inside of prepare, because it is allowed now to have functions nested like that, it will only throw in prepare (creating a deployment), not parse (uploading a blueprint)

so thats a change in behaviour in this pr

dsl_parser/tests/test_get_attribute.py
tehasdf: refactored, however note that this wasn't added in this pr

dsl_parser/tests/test_register_function.py
tehasdf: refactored, however note that this wasn't added in this pr

dsl_parser/tests/test_functions.py
tehasdf: ditto previous comment, re-added/changed a bit preexisting tests

dsl_parser/functions.py
tehasdf: what can be more clear than this? it's messy, gotta clean it up, the jira is about cleaning it up

cloudify/workflows/local.py
tehasdf: the nodes are handed out to plugins, which can mutate them, so they have to be deepcopied

cloudify/workflows/local.py
tehasdf: not sure about a lot - changed some things, lost 3? lines...

dsl_parser/functions.py
tehasdf: changed the docstring, however note that this wasn't added in this pr

dsl_parser/functions.py
tehasdf: i dont think so

in general this is ugly, and somewhat related to CY-1676 (cleaning up get-property)
this is tightly coupled to https://github.com/cloudify-cosmo/cloudify-common/blob/00f0abbf9f3d0168fe2aa61f44870965ad68ca88/cloudify/endpoint.py#L135-L142

an alternative would be to probably have an explicit `type` field, or have that other piece of code call a more specific thing than just the generic "evaluate functions" endpoint

dsl_parser/functions.py
```diff
@@ -892,70 +848,113 @@ def handler(v, scope, context, path):
...
-                               get_capability_method):
-    return _handler('evaluate_runtime',
-                    storage=RuntimeEvaluationStorage(
-                        get_node_instances_method=get_node_instances_method,
-                        get_node_instance_method=get_node_instance_method,
-                        get_node_method=get_node_method,
-                        get_secret_method=get_secret_method,
-                        get_capability_method=get_capability_method))
+    def get_node(self, node_name):
+        for n in self._plan.node_templates:
```
tyacbovi: maybe move it to list comprehension for and return the first

dsl_parser/tests/test_functions.py
```diff
@@ -719,6 +718,55 @@ def get_property_from_get_property(self):
         plan = prepare_deployment_plan(self.parse(yaml))
         self.assertEqual('secret', plan['nodes'][0]['properties']['b'])
 
+    def test_get_property_runtime(self):
```
tyacbovi: maybe add tests that runs get_prop that had inside get_secret and see that also works with the flag

dsl_parser/tests/test_functions.py
```diff
@@ -1251,109 +1299,99 @@ def test_func_integrations_within_list(self):
...
         with self.assertRaisesRegexp(
-                exceptions.FunctionValidationError,
-                'Runtime function .+ cannot be nested within a non-runtime '
-                'function \\(found in .+\\)'):
+                exceptions.EvaluationRecursionLimitReached,
+                "The recursion limit \\([0-9]+\\) has been reached while "
+                "evaluating the deployment\\..+get_property\\[0\\].+"):
             prepare_deployment_plan(self.parse(yaml))
 
+    def test_get_input_doesnt_accept_runtime_func_as_args(self):
```
tyacbovi: well I am still for multi test functions of quicker understanding of what's wrong when it will fail

dsl_parser/tests/test_functions.py
```diff
@@ -1366,107 +1404,112 @@ def test_get_property_doesnt_accept_runtime_func_as_args(self):
...
                 get_input:
-                    get_property:
-                        - some_node
-                        - prop
+                    concat:
+                        - get_attribute:
+                                - some_node
+                                - dummy_prop
     """
+        plan = self.parse_1_1(yaml)
```
tyacbovi: why use parse_1_1?

dsl_parser/tests/test_outputs.py
```diff
@@ -138,7 +138,8 @@ def test_valid_get_secret(self):
         func = functions.parse(outputs['port']['value'])
         self.assertTrue(isinstance(func, functions.GetSecret))
```
tyacbovi: There are no changes in capabilities test?

dsl_parser/tests/test_functions.py
tyacbovi: also get prop running a get input

cloudify_rest_client/deployment_updates.py
```diff
@@ -208,15 +210,18 @@ def update(self,
             params['skip_install'] = skip_install
         if skip_uninstall:
             params['skip_uninstall'] = skip_uninstall
-        if skip_reinstall:
+        if skip_reinstall is not None:
```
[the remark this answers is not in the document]
tehasdf: otherwise you can't pass False
bug unrelated to this pr but might as well fix it

dsl_parser/functions.py
tehasdf: handlers are now free to implement a missing input however they wish, from a keyerror, or a unknown input error, or hell, it can even pop an input to ask the operator

dsl_parser/functions.py
tehasdf: me too but that's what CY-1676 is about, not this issue :P

cloudify/workflows/local.py
```diff
@@ -424,25 +416,33 @@ def update_node_instance(self,
                 instance['state'] = state
             self._store_instance(instance)
 
-    def _get_node_instance(self, node_instance_id):
+    def get_node_instance(self, node_instance_id, evaluate_functions=True):
```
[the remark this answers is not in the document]
tehasdf: doesn't really matter for cfy local but disabled

dsl_parser/functions.py
```diff
@@ -772,115 +749,94 @@ def parse(raw_function, scope=None, context=None, path=None):
...
-        :return: relevant string containing information about the arguments.
-        """
-        msg = "Limit was reached with the following path - {0}"
-        if args and len(args) == 4:
-            return msg.format(args[3])
-        if kwargs and 'path' in kwargs:
-            return msg.format(kwargs['path'])
-        return ""
+    """
+    return "Limit was reached with the following path - {0}".format(path)
```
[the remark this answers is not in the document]
tehasdf: no, it's still used :(

dsl_parser/functions.py
tehasdf: that'd just use more memory unnecessarily (for the list)

dsl_parser/tests/test_functions.py
tehasdf: that's what the original test used, but i now made sure that it's easier to see in the commit diff

cloudify/workflows/workflow_context.py
```diff
@@ -966,8 +966,9 @@ def __init__(self, ctx):
                 raw_node_instances = storage.get_node_instances()
             else:
                 rest = get_rest_client()
-                raw_nodes = rest.nodes.list(deployment_id=self.deployment.id,
-                                            _get_all_results=True)
+                raw_nodes = rest.nodes.list(
+                    deployment_id=self.deployment.id, _get_all_results=True,
```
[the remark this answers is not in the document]
tehasdf: OK, but plz no style comments based on personal preference (not in flake)

dsl_parser/functions.py
```diff
@@ -228,14 +226,21 @@ def validate(self, plan):
...
-                "get_input function references an "
-                "unknown input '{0}'.".format(self.input_value))
-        return plan.inputs[self.input_value]
+            if any(is_function(x) for x in self.input_value):
+                raise exceptions.FunctionEvaluationError(
+                    '{0}: found an unresolved argument: {1}'
+                    .format(self.name, self.input_value))
+
+            return self._get_input_attribute(
+                handler.get_input(self.input_value[0]))
```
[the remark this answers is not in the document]
tehasdf: there is more but the first one is the input name
the rest is handled inside of `self._get_input_attribute`
(this logic is unchanged)

dsl_parser/tests/test_outputs.py
tehasdf: probably no reason, get_capability is a runtime func anyway

dsl_parser/functions.py
tehasdf: there isn't an evaluation handler existing at this point yet, so we can't pass it in

it's very ugly that the validate method needs to evaluate, but this is what it did before, and i dont want to change it in this pr

dsl_parser/functions.py
```diff
@@ -302,9 +303,10 @@ def validate(self, plan):
         args = [self.node_name] + self.property_path
         if any(is_function(x) for x in args):
             return
-        self.evaluate(plan)
+        handler = _PlanEvaluationHandler(plan, runtime_only_evaluation=False)
```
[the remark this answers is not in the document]
tehasdf: validate is always parse-time, we can't use runtime stuff at validation time

PR #359, opened 2019-09-04
CY-1710 restclient: Add `update_executions` to dep-update

tehasdf:


cloudify_rest_client/deployment_updates.py
```diff
@@ -208,15 +208,14 @@ def update(self,
             params['skip_install'] = skip_install
         if skip_uninstall:
             params['skip_uninstall'] = skip_uninstall
-        if skip_reinstall:
+        if skip_reinstall is not None:
```
tehasdf: not really related to the original change, but with just a `if skip_reinstall:` we're not able to pass in a `False`

tehasdf:
not needed because CY-1710 ended up doing this by default instead of by a flag

PR #361, opened 2019-09-08
CY-1711 Don't duplicate context in serialized tasks

tehasdf:
cloudify_context is already present in the task kwargs, so no need
to store it there twice.
However it is required as a parameter to __init__, and so in restore
we must make sure to pass it there.


cloudify/tests/test_task_serialize.py
```diff
@@ -21,9 +21,12 @@
 
 
 def _make_remote_task(kwargs=None):
+    if kwargs is None:
```
tyacbovi: you still can do this 'kwargs=kwargs or {'a': 1}'

PR #365, opened 2019-09-26
CY-1374 Basic py2+3 compat changes

tehasdf:
Just "stage1" changes that carry no risk and are 2+3 safe.
This will allow importing some of the modules under py3, which
will allow some users to start porting their plugins to py3

### DIFF
--- a/cloudify/tests/test_context.py
+++ b/cloudify/tests/test_context.py
@@ -668,7 +668,6 @@ def get_template(ctx, testing, **_):
                 rendering_tests_demo_conf,
                 template_variables={'ctx': ctx})
         except exceptions.NonRecoverableError:
-            print 'caught expected exception'
             resource = 'failed'
 
     if testing == 'get_resource_no_template':
@@ -698,7 +697,6 @@ def download_template(ctx, testing, **_):
                 rendering_tests_demo_conf,
                 template_variables={'ctx': ctx})
         except exceptions.NonRecoverableError:
-            print 'caught expected exception'
             resource = 'failed'
 
     if testing == 'download_resource_no_template':
--- a/docs/cloudify_rest_client/conf.py
+++ b/docs/cloudify_rest_client/conf.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 # flake8: NOQA
 # -*- coding: utf-8 -*-
 #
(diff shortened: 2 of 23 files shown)


cloudify/tests/test_context.py
```diff
@@ -668,7 +668,6 @@ def get_template(ctx, testing, **_):
                 rendering_tests_demo_conf,
                 template_variables={'ctx': ctx})
         except exceptions.NonRecoverableError:
-            print 'caught expected exception'
```
geokala: Shouldn't we leave the print in, just as a function?

geokala: (ditto for other prints)

docs/cloudify_script_plugin/conf.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import print_function
```
geokala: This isn't actually necessary for the prints we're doing, we can just use the function syntax but not the function itself (unless that breaks on 2.6?)

dsl_parser/constraints.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
```
geokala: Do you really need to import absolute_import, or can you just make the imports actually absolute?

cloudify/tests/test_context.py
tehasdf: no, those prints do nothing, they're in the tests and there's mostly no way to recover the log that those prints go into anyway; if someone needs logs there (which they tbh shouldn't) they are welcome to add a ctx.logger call instead

docs/cloudify_script_plugin/conf.py
tehasdf: it's not really the function syntax then, it's printing a string which is wrapped in parens for no reason; I'm not gonna do this because I don't like code which does one thing but looks like it does another :P

geokala: Fair enough, marking resolved.

cloudify/tests/test_context.py
geokala: Sounds fair, marking resolved.

dsl_parser/constraints.py
tehasdf: yeah good point, this isnt needed, removed

the explicit-relative imports would work just fine without this `__future__`, but i've made them absolute anyway as well

docs/cloudify_rest_client/conf.py
```diff
@@ -73,13 +74,13 @@
 try:
     release = pkg_resources.get_distribution(project).version
 except pkg_resources.DistributionNotFound:
-    print 'To build the documentation, The distribution information of packm'
-    print 'Has to be available.  Either install the package into your'
-    print 'development environment or run "setup.py develop" to setup the'
-    print 'metadata.  A virtualenv is recommended!'
+    print('To build the documentation, The distribution information of packm')
```
tyacbovi: This is not our code, it's sphinx.. So maybe there is a new version of this

PR #332, opened 2019-07-01
CY-1374 Python 3 support

tehasdf:
- [x]  apply 2to3 fix-by-fix and tune the outputs maually
- [x] run unittests on py3
- [x] make flake pass on py3 (should be immediate after 2to3 is done)
- [x] make the tests pass on py2 again
- [ ] make the tests pass on py3

PR #366, opened 2019-09-27
CY-1374 Move several imports to a compat module

tehasdf:
For 2+3 compatibility.
This actually allows `import cloudify.mocks` in py3


tehasdf:
this is pretty much factored out of #332

PR #370, opened 2019-10-27
Cy 1750 node instance serialization decorator

EarthmanT:


cloudify/utils.py
```diff
@@ -266,6 +266,45 @@ def get_execution_token():
...
 
+def get_instances_of_node(deployment_id=None, node_id=None, _ctx=None):
+    """
+    Get node instances of a node from a deployment.
+    """
+    _ctx = _ctx or _get_current_context()
+    deployment_id = deployment_id or _ctx.deployment.id
+    node_id = node_id or _ctx.node.id
+    from cloudify.manager import get_rest_client
+    rest_client = get_rest_client()
```
mabuaisha: @EarthmanT what about passing the rest client as part of the method signature instead of using import inside the body of this method ?, so I'm pretty sure that rest client is importing the `utils` and the way you put the import here in order to avoid the circular dependency

cloudify/decorators.py
```diff
@@ -40,4 +45,87 @@ def workflow(func=None, system_wide=False, resumable=False, **kwargs):
...
+    :return: decorator wrapper
+    """
+
+    workflows = workflows if isinstance(
+        workflows, list) else ['install']
+    states = states if isinstance(
+        states, list) else ['started']
+    rest_client = get_rest_client()
+
+    def outer_wrapper(func):
```
tehasdf: needs `@functools.wraps`

cloudify/utils.py
```diff
@@ -266,6 +266,50 @@ def get_execution_token():
...
+    Get node instances of a node from a deployment.
+    """
+    _ctx = _ctx or _get_current_context()
+    deployment_id = deployment_id or _ctx.deployment.id
+    node_id = node_id or _ctx.node.id
+    return rest_client.node_instances.list(deployment_id=deployment_id,
+                                           node_id=node_id)
+
+
+def split_instances_at_index(rest_client,
```
tehasdf: this function does A LOT, wouldn't it be better if it was just `def split_instances_at_index(instances, index)`? I don't think this is the best place for actually fetching the node instances

cloudify/utils.py
```diff
@@ -266,6 +266,50 @@ def get_execution_token():
...
+                             deployment_id=None,
+                             node_id=None,
+                             current_node_instance=None,
+                             index_delta=None,
```
tehasdf: do we really need `index_delta`?

cloudify/utils.py
```diff
@@ -266,6 +266,50 @@ def get_execution_token():
...
+    Create and sort the list of instances of a node. Then split the lists
+    at the delta from the index of the current node instance.
+    """
+    current_node_instance = current_node_instance or ctx.instance
+    index_delta = index_delta or 1
+    _ctx = _ctx or _get_current_context()
+    instances_in_queue = get_instances_of_node(
+        rest_client, deployment_id, node_id, _ctx)
+    # Sort instances by index field.
+    instances_in_queue.sort(key=lambda n: n.index)
```
tehasdf: the sorting isn't actually required for anything

cloudify/decorators.py
```diff
@@ -40,4 +45,87 @@ def workflow(func=None, system_wide=False, resumable=False, **kwargs):
...
+                            preceding_instance and \
+                            preceding_instance.state not in states:
+                        raise CloudifySerializationRetry(
+                            'Serial Cloudify Operation. '
+                            'Waiting for the following '
+                            'node instances to be ready: '
+                            '{priority_instances}.'.format(
+                                priority_instances=[(p.id, p.state) for p in
+                                                    priority_instances]))
+            return operation(func, *args, **kwargs)
```
tehasdf: this needs to eventually call the function
a decorator can be written like this

```python
def serial_operation(...):
  @functools.wraps
  def outer(f):
    def inner(*a, **kw):
      return f(*a, **kw)
    return operation(func=inner)
  return outer
```

tehasdf: also remember to pass the decorator arguments to `operation`'s kwargs (so that you can `@serial_operation(resumable=True)`

cloudify/tests/test_decorators.py
```diff
@@ -184,3 +188,110 @@ def w2():
...
+
+        # Check that we raise if the preceder is not 'started'.
+        _ctx._context['workflow_id'] = 'install'
+        current_ctx.set(_ctx)
+
+        _mock.side_effect = lambda: rest_client_mock.MockRestclient()
+
+        @serial_operation()
+        def run_op_install_only(**kwargs):
+            run(assert_op_impl, **kwargs)
```
tehasdf: this line (and other operations) doesn't run because the decorator is broken

cloudify/tests/test_decorators.py
```diff
@@ -184,3 +188,110 @@ def w2():
...
+        @serial_operation(states=['uninitialized'])
+        def run_op_uninitialized_states(**kwargs):
+            run(assert_op_impl, **kwargs)
+
+        @serial_operation(workflows=['scale'])
+        def run_op_neither(**kwargs):
+            run(assert_op_impl, **kwargs)
+
+        list_fn.return_value = [
+            MagicMock(id='test_node{x}'.format(x=n),
```
tehasdf: could well be just `Mock`, no reason this has to be `MagicMock`

cloudify/decorators.py
```diff
@@ -40,4 +45,87 @@ def workflow(func=None, system_wide=False, resumable=False, **kwargs):
...
+                # current node instances's. They should be executed first.
+                # "no_priority" means that it comes after.
+                priority_instances, no_priority = \
+                    split_instances_at_index(rest_client)
+                # Finished instances are those that are already in the
+                # state where they are not considered blockers of our
+                # current node instance.
+                non_blocking_instances = [
+                    i for i in priority_instances + no_priority
+                    if i.state in states]
```
tehasdf: i'd refactor this to return early right here, if we already hit the threshold, because then we would be able to forget about the threshold after this, and the big `if` condition would've been less crazy
(which would also imply factoring the logic out to a separate `def should_wait` function)

cloudify/decorators.py
```diff
@@ -40,4 +45,87 @@ def workflow(func=None, system_wide=False, resumable=False, **kwargs):
...
+                    #    This means - always try to execute pure serial.
+                    # 2) revert_to_parallel_threshold is unequal to total
+                    #    finished instances. This means that if we have
+                    #    exceeded the revert_to_parallel_threshold threshold,
+                    #    then we would return to parallel execution.
+                    if (revert_to_parallel_threshold == 0 or
+                        revert_to_parallel_threshold != len(
+                                non_blocking_instances)) and \
+                            preceding_instance and \
+                            preceding_instance.state not in states:
```
tehasdf: oh, i understand - this actually requires the sort, because we want to find the very previous instance to check its state

cloudify/decorators.py
```diff
@@ -40,4 +45,87 @@ def workflow(func=None, system_wide=False, resumable=False, **kwargs):
...
+        The default is ['started']. We will only continue to next instance
+        after the preceding instance is in "started" state.
+    :raises: CloudifySerializationRetry.
+    :return: decorator wrapper
+    """
+
+    workflows = workflows if isinstance(
+        workflows, list) else ['install']
+    states = states if isinstance(
+        states, list) else ['started']
```
tehasdf: just ignoring the passed-in value if it's of the wrong type, is pretty bad api

cloudify/decorators.py
tehasdf: then maybe write this more directly, without splitting, and just find the preceding instance using a forloop or otherwise
(eg. `preceding_instance = next((ni for ni in instances if ni.index == current_ni.index - 1), None)` or `preceding = None;  for ni in instances:  if ni.index == current_ni.index - 1: preceding = ni; break`)

cloudify/decorators.py
```diff
@@ -40,4 +45,87 @@ def workflow(func=None, system_wide=False, resumable=False, **kwargs):
...
+                    # We want to wait if the following conditions exist:
+                    # 1) revert_to_parallel_threshold is 0 and preceding
+                    #    node instance (preceding_instance) is not ready.
+                    #    This means - always try to execute pure serial.
+                    # 2) revert_to_parallel_threshold is unequal to total
+                    #    finished instances. This means that if we have
+                    #    exceeded the revert_to_parallel_threshold threshold,
+                    #    then we would return to parallel execution.
+                    if (revert_to_parallel_threshold == 0 or
+                        revert_to_parallel_threshold != len(
```
tehasdf: as a threshold, this has to be `>`/`>=`, not `!=`?

cloudify/decorators.py
```diff
@@ -40,4 +45,87 @@ def workflow(func=None, system_wide=False, resumable=False, **kwargs):
...
+                                non_blocking_instances)) and \
+                            preceding_instance and \
+                            preceding_instance.state not in states:
+                        raise CloudifySerializationRetry(
```
tehasdf: will this count for operation retries, and could possibly exhaust the limit?

cloudify/decorators.py
EarthmanT: will change it.

cloudify/decorators.py
EarthmanT: will do

cloudify/tests/test_decorators.py
EarthmanT: OK

cloudify/decorators.py
EarthmanT: Will add code to handle a limit.

cloudify/decorators.py
EarthmanT: You mean I need to add that in the serial_operation args?

PR #372, opened 2019-10-29
CY-1534 Add inter-deployment dependency infrastructure

davidginzbourg:


dsl_parser/functions.py
```diff
@@ -588,22 +597,88 @@ def __init__(self, args, **kwargs):
...
-        pass
+        if plan.version.definitions_version < (1, 1):
+            raise exceptions.FunctionValidationError(
+                'Using {0} requires using dsl version 1_1 or '
+                'greater, but found: {1} in {2}.'
+                .format(self.name, plan.version, self.path))
 
     def evaluate(self, handler):
-        return handler.get_secret(self.secret_id)
+        for joined_value in self.joined:
```
tyacbovi: not clear naming, not sure the logic here

dsl_parser/functions.py
```diff
@@ -588,22 +597,88 @@ def __init__(self, args, **kwargs):
...
                 "`get_secret` function argument should be a string\\dict "
                 "(a function). Instead it is a {0} with the "
                 "value: {1}.".format(type(args), args))
         self.secret_id = args
 
+    def evaluate(self, handler):
+        return handler.get_secret(self.secret_id)
+
+
+@register(name='concat', func_eval_type=HYBRID_FUNC)
```
tyacbovi: let's go over the file f2f

tehasdf:
BTW when creating a pull request you can click the "create a draft" button and then it will not be mergeable until you mark it ready, see https://github.blog/2019-02-14-introducing-draft-pull-requests/

dsl_parser/scan.py
```diff
@@ -81,40 +89,44 @@ def _scan_operations(operations,
...
                      replace=replace)
     for r in node_template.get('relationships', []):
         context = {'node_template': node_template, 'relationship': r}
         _scan_operations(r.get('source_operations', {}),
                          handler,
                          scope=NODE_TEMPLATE_RELATIONSHIP_SCOPE,
                          context=context,
-                         path='{0}.{1}'.format(node_template['name'],
-                                               r['type']),
+                         path='{0}.{1}.{2}'.format(NODES,
```
tehasdf: tbh those changes i'd prefer to see as a separate PR, because i don't think this is related to inter-deployment dependencies

dsl_parser/functions.py
tehasdf: this is concat, so joins values, so those are joined
anyway, this isn't changed in this PR, (used to be the same before), only git marks it as added+removed because concat was moved up

@davidginzbourg perhaps you can move it down so that git doesn't mark those lines as changed

anyway, this is fiiiine

cloudify_rest_client/inter_deployment_dependencies.py
```diff
@@ -0,0 +1,153 @@
...
+            _include=None,
+            **kwargs):
+        """
+        Gets an inter-deployment dependency by its id.
+
+        :param dependency_id: InterDeploymentDependency's id to get.
+        :param _include: List of fields to include in response.
+        :return: The inter-deployment dependency details.
+        """
+        assert dependency_creator
```
tehasdf: plz no additional `assert`s, i do realize we have some of them, but it's pretty bad, because as you know asserts are entirely removed in "optimized mode" (`python -O`), but we do need those checks

insetad, please do something like
```python
if not (dependency_creator and source_deployment and target_deployment):
    raise ValueError('blabla')
```
feel free to write the `and` logic differently (3 separate `if`s are fine), and perhaps use CloudifyClientError, but a ValueError (or even RuntimeError) are better than asserts

cloudify_rest_client/inter_deployment_dependencies.py
```diff
@@ -0,0 +1,153 @@
...
+         depends on.
+        :param doesnt_exist_ok: if it's False then no error will be raised if
+         the dependency doesn't exist. Defaults to False.
+        :return: an InterDeploymentDependency object.
+        """
+        data = {
+            'dependency_creator': dependency_creator,
+            'source_deployment': source_deployment,
+            'target_deployment': target_deployment,
+            'doesnt_exist_ok': doesnt_exist_ok
```
tehasdf: `doesnt_exist_ok` feels like something that should be handled here, on the client-side, instead of on the server-side

tehasdf: would be a try/except, but however we do have the concept of `.get(..., expected_status_code=x)`, but it has to be a single specific status code
perhaps we should change it to allow expecting a list of codes

cloudify/utils.py
```diff
@@ -51,6 +51,8 @@
 INSPECT_TIMEOUT = 30
 ADMIN_API_TOKEN_PATH = '/opt/mgmtworker/work/admin_token'
 
+string_types = (basestring,) if sys.version_info[0] == 2 else (str, unicode)
```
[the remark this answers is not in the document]
tyacbovi: yes, agree.
Maybe a new package of compat, because there is no use of dup code

YoniItzhak:
This PR is replaced by [CY-1534 component and shared resource infrastructure](https://github.com/cloudify-cosmo/cloudify-common/pull/458)

dsl_parser/functions.py
```diff
@@ -605,8 +612,76 @@ def evaluate(self, handler):
...
+class InterDeploymentDependencyCreatingFunction(Function):
+
+    def register_function_to_plan(self, plan):
+        if not self.function_identifier:
+            return
+
+        plan.setdefault(
+            INTER_DEPLOYMENT_FUNCTIONS,
+            {}
+        )[self.function_identifier] = self.target_deployment
```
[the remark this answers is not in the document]
YoniItzhak: This point was resolved in the discussion we had, where we decided to keep both.

dsl_parser/functions.py
```diff
@@ -637,44 +712,16 @@ def parse_args(self, args):
...
-        return self.join()
+        pass
 
-    def join(self):
-        str_join = [str(elem) for elem in self.joined]
-        return self.separator.join(str_join)
+    @property
+    def target_deployment(self):
+        first_arg = self.capability_path[0]
+        return None if is_function(first_arg) else first_arg
```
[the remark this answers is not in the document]
YoniItzhak: What we would have liked to check more?

PR #381, opened 2019-12-03
CY-1905 Add constants

davidginzbourg:


cloudify/constants.py
```diff
@@ -95,3 +95,9 @@
 CLOUDIFY_TOKEN_AUTHENTICATION_HEADER = 'Authentication-Token'
 
 NAMESPACE_BLUEPRINT_IMPORT_DELIMITER = '--'
+MANAGER_STATUS_REPORTER = 'manager_status_reporter'
```
tyacbovi: please move to cluster_status.py

PR #384, opened 2019-12-11
CY-1821 Add the snapshot restore marker file path constant

davidginzbourg:


cloudify/snapshots.py
```diff
@@ -0,0 +1,25 @@
...
+#    * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    * See the License for the specific language governing permissions and
+#    * limitations under the License.
+
+from collections import namedtuple
+
+SNAPSHOT_RESTORE_FLAG_FILE = ('/opt/cloudify/snapshot_status'
+                              '/snapshot_restoring')
+
+SnapshotRestoreStates = namedtuple('SnapshotRestoreStates',
```
tyacbovi: You can this with a simple class, and it will look better

PR #388, opened 2019-12-25
Cy 1922 add cloudify versions to plugin declaration

mabuaisha:


mabuaisha:
> @mabuaisha Do we still want this? If we're not sure we do, let's close it. If we do want it, let's get it into 6.1.

I think this is needed since there is a jira for that but not sure when should we have this

PR #389, opened 2019-12-30
CY-2150 Don't store tokens in operation parameters

tehasdf:


cloudify/workflows/tasks.py
```diff
@@ -420,11 +420,27 @@ def restore(cls, ctx, graph, task_descr):
...
+        # resumed, but we'll use the up-to-date token from workflow ctx
+        context['execution_token'] = ctx.execution_token
         return super(RemoteWorkflowTask, cls).restore(ctx, graph, task_descr)
 
     def dump(self):
         task = super(RemoteWorkflowTask, self).dump()
-        task['parameters']['task_kwargs'] = {'kwargs': self._kwargs}
+
+        # store a copy of kwargs but without __cloudify_context
+        task['parameters']['task_kwargs'] = {
```
inbala: Why not just check `if '__cloudify_context ' in self._kwargs`? pop it and save it?

tehasdf: we need to avoid mutating kwargs, because those values are not copied; thats why i'm using this, which is effectively a copy but with dropping the one field

we dont want a copy/deepcopy because that runs very often and copying the (large) context dict starts to affect perf :/

PR #391, opened 2019-12-31
CY-2153 Enable giving a request timeout when creating a client

inbala:


cloudify_rest_client/client.py
```diff
@@ -78,12 +78,13 @@ def __init__(self, host, port=DEFAULT_PORT,
...
                  headers=None, query_params=None, cert=None, trust_all=False,
                  username=None, password=None, token=None, tenant=None,
-                 kerberos_env=None):
+                 kerberos_env=None, timeout=None):
         self.port = port
         self.host = host
         self.protocol = protocol
         self.api_version = api_version
         self.kerberos_env = kerberos_env
+        self.default_timeout_sec = timeout
```
tyacbovi: This should be 'timeout or self.default_timeout_sec'

cloudify/cluster.py
```diff
@@ -38,6 +35,9 @@ def __init__(self, *args, **kwargs):
         random.shuffle(hosts)
         self.hosts = itertools.cycle(hosts)
         self.host = hosts[0]
+        self.default_timeout_sec = (5, None)
```
tyacbovi: why not pass via the super's init?

inbala: It defines also the related values of
```
self.retries = 30
self.retry_interval = 3
```

PR #393, opened 2020-01-03
CY-1960 Creating queue: add x-ha-policy

tehasdf:


geokala:
This is covered by the policies which are now copied to new vhosts, so is not required.

PR #398, opened 2020-01-14
CY-2173: adding status reporters user ids

tyacbovi:


cloudify/cluster_status.py
```diff
@@ -1,5 +1,5 @@
 ########
-# Copyright (c) 2019 Cloudify Platform Ltd. All rights reserved
+# Copyright (c) 2019-2020 Cloudify Platform Ltd. All rights reserved
```
tehasdf: that seems odd

PR #399, opened 2020-01-15
CY-1959 If operation was started before, mark it as a resume

tehasdf:
When setting operation state to "started", we might learn that the
operation was already started before. In that case, we know it's a
resume, so we can set it as one.


tehasdf:
cloudify-cosmo/cloudify-manager#2084 is the required api change

PR #404, opened 2020-02-11
Update when to release sphinx docs and increase the version number to…

mabuaisha:
… 4 digits


.circleci/config.yml
```diff
@@ -157,10 +157,10 @@ workflows:
               - docs_plugins_common
           filters:
             branches:
-              only: master
+              only: /([0-9\.]*\-build)/
```
tyacbovi: Notice that if someone will push a change to an older release branch, that will overwrite the more recent one

tehasdf: yeah we really need docs with a version select dropdown
make a jira? @mabuaisha

mabuaisha: @tyacbovi I think for know we can pin it to 5.0.5-build and maybe later on we can update it to dupport multiple versions

PR #408, opened 2020-02-24
CY-2359 Python3: futurize stage 1 and CI

tehasdf:
This applies most of futurize --stage1, and creates a CI to check
that no incompatible code is added during the transitory period.


tehasdf:
This introduces a new CI check, `py3_compat`.

The check looks into the list in .circleci/py3fixers, and verifies that no code in the repository is incompatible with the fixers already applied.

For a demonstration of the check failing, see #409

PR #410, opened 2020-02-24
CY-2367 python3 porting: stage2 changes

tehasdf:
See commits

### DIFF
--- a/cloudify/snapshots.py
+++ b/cloudify/snapshots.py
@@ -13,10 +13,11 @@
 #    * See the License for the specific language governing permissions and
 #    * limitations under the License.
 
+from builtins import object
 SNAPSHOT_RESTORE_FLAG_FILE = ('/opt/manager/snapshot_status'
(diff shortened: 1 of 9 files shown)


cloudify/snapshots.py
```diff
@@ -13,10 +13,11 @@
 #    * See the License for the specific language governing permissions and
 #    * limitations under the License.
 
+from builtins import object
 SNAPSHOT_RESTORE_FLAG_FILE = ('/opt/manager/snapshot_status'
                               '/snapshot_restoring')
 
 
-class STATES:
+class STATES(object):
```
geokala: Is this inheritance actually required, or is it just to quiet one of the complaining futurizers?

tehasdf: just for the futurizer; in py2, an old-style class would work in this case (but would stop working as soon as we tried to use eg. `@property`)

PR #415, opened 2020-02-28
CY-2367 Py3 porting: step6

tehasdf:
Apply the step6 fixers:
```
lib2to3.fixes.fix_filter
lib2to3.fixes.fix_dict
lib2to3.fixes.fix_map
lib2to3.fixes.fix_zip
lib2to3.fixes.fix_xrange
```

See commit by commit, and especially see the notes in the fix_dict commit

### DIFF
--- a/cloudify/tests/test_local_workflows.py
+++ b/cloudify/tests/test_local_workflows.py
@@ -24,7 +24,6 @@ import threading
 import Queue
 
 import testtools
-from testtools.matchers import ContainsAll
 
 import cloudify.logs
 from cloudify.decorators import workflow, operation
@@ -615,10 +614,8 @@ class LocalWorkflowTest(BaseWorkflowTest):
             self.assertEqual(['type'], node1.type_hierarchy)
             self.assertEqual(['type', 'cloudify.nodes.Compute'],
                              node2.type_hierarchy)
-            self.assertThat(node1.properties.items(),
-                            ContainsAll({'property': 'value'}.items()))
-            self.assertThat(node2.properties.items(),
-                            ContainsAll({'property': 'default'}.items()))
+            self.assertEqual(node1.properties.get('property'), 'value')
+            self.assertEqual(node2.properties.get('property'), 'default')
             self.assertEqual(sorted_ops, sorted(node1.operations.keys()))
             self.assertEqual(sorted_ops, sorted(node2.operations.keys()))
             self.assertIs(relationship, node1.get_relationship('node2'))
@@ -660,7 +657,7 @@ class LocalWorkflowTest(BaseWorkflowTest):
         def op1(ctx, **_):
             caps = ctx.capabilities.get_all()
             self.assertEqual(1, len(caps))
-            key, value = next(caps.iteritems())
+            key, value = dict(caps).popitem()
             self.assertIn('node2_', key)
             self.assertEqual(value, {'key': 'value'})
 
@@ -690,11 +687,11 @@ class LocalWorkflowTest(BaseWorkflowTest):
 
         def op(ctx, **_):
             if 'node2_' in ctx.target.instance.id:
-                self.assertThat(ctx.target.node.properties.items(),
-                                ContainsAll({'property': 'default'}.items()))
+                self.assertEqual(
+                    ctx.target.node.properties.get('property'), 'default')
             elif 'node_' in ctx.target.instance.id:
-                self.assertThat(ctx.target.node.properties.items(),
-                                ContainsAll({'property': 'value'}.items()))
+                self.assertEqual(
+                    ctx.target.node.properties.get('property'), 'value')
             else:
                 self.fail('unexpected: {0}'.format(ctx.target.instance.id))
         self._execute_workflow(the_workflow, operation_methods=[op])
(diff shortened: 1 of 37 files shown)


cloudify/tests/test_local_workflows.py
```diff
@@ -615,10 +614,8 @@ def blueprint_model(ctx, **_):
             self.assertEqual(['type'], node1.type_hierarchy)
             self.assertEqual(['type', 'cloudify.nodes.Compute'],
                              node2.type_hierarchy)
-            self.assertThat(node1.properties.items(),
-                            ContainsAll({'property': 'value'}.items()))
-            self.assertThat(node2.properties.items(),
-                            ContainsAll({'property': 'default'}.items()))
+            self.assertEqual(node1.properties.get('property'), 'value')
+            self.assertEqual(node2.properties.get('property'), 'default')
```
tehasdf: imagine doing 2 calls to `.items()` and a testtools `ContainsAll` to check one single damn value

cloudify/workflows/workflow_context.py
```diff
@@ -306,7 +306,7 @@ def node_id(self):
     @property
     def relationships(self):
         """The node relationships"""
-        return self._relationship_instances.itervalues()
+        return iter(self._relationship_instances.values())
```
tehasdf: those need to stay iterators because unfortunately things use this as iterators by calling `next`.
It would be better if usage was to just `for x in ..` instead of an explicit `next()`, but for now, let's keep back-compat

cloudify/workflows/workflow_context.py
```diff
@@ -1622,17 +1622,27 @@ def __init__(self, workflow_ctx, modification):
...
-            (instance.node_id, workflow_ctx.get_node(instance.node_id)._node)
-            for instance in node_instances.added_and_related).values()
+        added_raw_nodes = []
+        seen_ids = set()
+        for instance in node_instances.added_and_related:
+            if instance.node_id not in seen_ids:
+                added_raw_nodes.append(
+                    workflow_ctx.get_node(instance.node_id)._node)
+                seen_ids.add(instance.node_id)
+
```
tehasdf: this (+below) could've been simpler
```python
added_raw_nodes = [
    workflow_ctx.get_node(instance.node_id)._node
    for instance in node_instances.added_and_related
]
```
however the old code also did deduplication by node id, so we're going to do it too, but more explicitly (you can now actually see which items have precedence)

dsl_parser/elements/node_templates.py
```diff
@@ -537,7 +537,7 @@ def parse(self, host_types, plugins):
             processed_nodes=processed_nodes,
             host_types=host_types,
             plugins=plugins)
-        return processed_nodes.values()
+        return list(processed_nodes.values())
```
tehasdf: things like this are returned to who-knows-where, so let's keep backwards compat by still returning a list on py3

dsl_parser/elements/node_templates.py
```diff
@@ -593,7 +593,8 @@ def _process_nodes_plugins(processed_nodes,
                         if plugin[constants.PLUGIN_EXECUTOR_KEY] \
                                 == constants.HOST_AGENT:
                             plugins_to_install[plugin['name']] = plugin
-            node[constants.PLUGINS_TO_INSTALL] = plugins_to_install.values()
+            node[constants.PLUGINS_TO_INSTALL] = \
+                list(plugins_to_install.values())
```
tehasdf: those possibly go into json, so they need to stay actual lists

dsl_parser/rel_graph.py
```diff
@@ -904,7 +904,7 @@ def _containing_groups(self, node_id):
             succ = graph.succ[node_id]
             if succ:
                 assert len(succ) == 1
-                node_id = succ.keys()[0]
+                node_id = list(succ.keys())[0]
```
tehasdf: didnt really want to do a `.pop` for just the key. Meh, this might as well just be this

dsl_parser/interfaces/utils.py
```diff
@@ -39,9 +39,8 @@ def merge_schema_and_instance_inputs(schema_inputs,
                                      instance_inputs):
 
     flattened_schema_inputs = utils.flatten_schema(schema_inputs)
-    merged_inputs = dict(
-            flattened_schema_inputs.items() +
-            instance_inputs.items())
+    merged_inputs = dict(flattened_schema_inputs)
+    merged_inputs.update(instance_inputs)
```
tehasdf: yes this is equivalent

script_runner/tests/test_ctx_wrapper.py
```diff
@@ -168,9 +168,9 @@ def test_get_all_node_properties(self):
                   'ctx.returns(value)')
         result = self._run(script)
         res_dict = ast.literal_eval(result)
-        self.assertIn('ip', res_dict.keys())
+        self.assertIn('ip', res_dict)
         self.assertEquals(res_dict['ip'], '1.1.1.1')
-        self.assertIn('key', res_dict.keys())
+        self.assertIn('key', res_dict)
```
tehasdf: imagine asserting `k in d.keys()` :D

PR #417, opened 2020-03-02
CY-2367 Py3 porting: step 7

tehasdf:
This applies `fix_basestring` and related changes

See commit messages for details

### DIFF
--- a/cloudify/_compat.py
+++ b/cloudify/_compat.py
@@ -29,6 +29,7 @@ if PY2:
 def reraise(exception_type, value, traceback):
     raise exception_type, value, traceback
 """)
+    text_type = unicode
 
 else:
     import queue
@@ -37,4 +38,7 @@ else:
     def reraise(exception_type, value, traceback):
         raise value.with_traceback(traceback)
 
-__all__ = ['PY2', 'queue', 'StringIO', 'reraise']
+    text_type = str
+
+
+__all__ = ['PY2', 'queue', 'StringIO', 'reraise', 'text_type']
--- a/cloudify/test_utils/local_workflow_decorator.py
+++ b/cloudify/test_utils/local_workflow_decorator.py
@@ -69,7 +69,7 @@ def _expand_dictionary(inputs, func_self, func_args, func_kwargs):
     func_to_call = None
     if callable(inputs):
         func_to_call = inputs
-    elif isinstance(inputs, basestring):
+    elif isinstance(inputs, str):
         if func_self is None:
             raise ValueError("You cannot supply 'string' "
                              "references to 'self' object in "
--- a/cloudify/tests/test_local_workflows.py
+++ b/cloudify/tests/test_local_workflows.py
@@ -1454,7 +1454,7 @@ class LocalWorkflowEnvironmentTest(BaseWorkflowTest, testtools.TestCase):
             self._execute_workflow(workflow_name='does_not_exist')
             self.fail()
         except ValueError as e:
-            self.assertIn("['workflow0']", e.message)
+            self.assertIn("[workflow0]", str(e))
 
     def test_getting_contained_elements(self):
         def check_subgraph(ctx, **_):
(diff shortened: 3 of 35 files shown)


dsl_parser/utils.py
```diff
@@ -21,15 +21,9 @@
 import sys
 import re
 
-try:
```
tehasdf: this is unrelated, but since i already made a compat module, lets just move this now

cloudify/tests/test_local_workflows.py
```diff
@@ -1454,7 +1454,7 @@ def test_no_workflow(self):
             self._execute_workflow(workflow_name='does_not_exist')
             self.fail()
         except ValueError as e:
-            self.assertIn("['workflow0']", e.message)
+            self.assertIn("[workflow0]", str(e))
```
tehasdf: changing `e.message` to `str(e)` is pretty much unrelated, but hey, less warnings from tests
(e.message is deprecated and throws a warning)

dsl_parser/framework/parser.py
```diff
@@ -74,6 +69,22 @@ def _traverse_schema(self, schema, list_nesting=0):
...
+        """Check that schema_type is valid for typechecking
+
+        schema_type needs to be something that can go into
+        isinstance() checks, ie. a class, type, or tuple of classes
+        and types.
+
+        Additionally, an empty schema is not allowed.
+        """
+        if not schema_type:
+            raise exceptions.DSLParsingSchemaAPIException(1)
```
tehasdf: `1`, everyone's favourite error

dsl_parser/yaml_loader.py
```diff
@@ -67,9 +67,9 @@ def construct_yaml_set(self, node):
         return self._holder(obj, node)
 
     def construct_yaml_str(self, node):
-        obj = SafeConstructor.construct_yaml_str(self, node)
+        obj = self.construct_scalar(node)
         try:
-            obj = str(obj)
+            obj.encode('ascii')
```
tehasdf: we still keep the assumption that blueprints have to be ascii
maybe we'll drop it soon

cloudify/_compat.py
```diff
@@ -29,6 +29,7 @@
 def reraise(exception_type, value, traceback):
     raise exception_type, value, traceback
 """)
+    text_type = unicode
```
tehasdf: in `cloudify._compat` we do use the stricter approach with `text_type = unicode`; only in dsl-parser, we have the more lenient `text_type = (unicode, str)` that the commit message talks about (and hopefully we'll drop it there as well soon)

cloudify/test_utils/local_workflow_decorator.py
```diff
@@ -69,7 +69,7 @@ def _expand_dictionary(inputs, func_self, func_args, func_kwargs):
     func_to_call = None
     if callable(inputs):
         func_to_call = inputs
-    elif isinstance(inputs, basestring):
+    elif isinstance(inputs, str):
```
geokala: Is there a reason this doesn't use text_type- just because it's a test or?

dsl_parser/tests/test_get_capability.py
```diff
@@ -216,7 +216,6 @@ def test_get_capability_not_list(self):
         self._assert_parsing_fails(
             yaml,
             message="`get_capability` function argument should be a list. "
-            "Instead it is a <type 'str'> with the value: i_should_be_a_list."
```
tehasdf: just, ugh, screw those error messages. I'll live with the slightly less tightly coupled asserts.

dsl_parser/yaml_loader.py
geokala: Is there any reason to keep this assumption, or just that we need to test before changing this?

geokala: From discussion: We'll leave this as-is now, because a change like this (while it shouldn't break anything) does need testing to make sure it actually accomplishes something useful rather than erroring out in an apparently unrelated place much deeper in the codebase.

cloudify/test_utils/local_workflow_decorator.py
tehasdf: eh, might as well leave it as "native str" here, it's just a test util, and used 1 time, too

dsl_parser/yaml_loader.py
tehasdf: no real reason but let's keep behaviour changes outside the scope of this

cloudify/test_utils/local_workflow_decorator.py
geokala: (yes- resolved)

PR #418, opened 2020-03-02
CY-2367 Py3 porting: more stage2 changes

tehasdf:
See commits

### DIFF
--- a/cloudify/ctx_wrappers/ctx-py.py
+++ b/cloudify/ctx_wrappers/ctx-py.py
@@ -6,6 +6,13 @@ import shlex
 import subprocess
 from collections import MutableMapping, Mapping
 
+PY2 = sys.version_info[0] == 2
+
+if PY2:
+    text_type = unicode
+else:
+    text_type = str
+
 
 def check_output(*popenargs, **kwargs):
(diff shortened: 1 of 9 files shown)


cloudify/ctx_wrappers/ctx-py.py
```diff
@@ -6,6 +6,13 @@
 import subprocess
 from collections import MutableMapping, Mapping
 
+PY2 = sys.version_info[0] == 2
+
+if PY2:
```
tehasdf: separate here, because `ctx-py.py` is supposed to be a standalone script (cannot import from cloudify modules)

cloudify/ctx_wrappers/ctx-py.py
```diff
@@ -34,7 +41,7 @@ def check_output(*popenargs, **kwargs):
 
 
 def unicode_to_string(text):
```
tehasdf: the whole premise of this function is wrong, because we should be dealing with unicode internally, and not bytes, so there shouldn't be a need for a dump-to-bytes util. But for now, i'm not making changes bigger than necessary

PR #420, opened 2020-03-02
CY-2363 Add a "validate blueprint" REST endpoint

mateumann:
Add `BlueprintsClient.validate()` method in `cloudify_rest_client` module for handling blueprints validation.  As validation is quite similar to upload, there is some refactoring as well.

### DIFF
--- a/cloudify_rest_client/blueprints.py
+++ b/cloudify_rest_client/blueprints.py
@@ -86,18 +86,15 @@ class BlueprintsClient(object):
         self._uri_prefix = 'blueprints'
         self._wrapper_cls = Blueprint
 
-    def _upload(self,
-                archive_location,
-                blueprint_id,
-                application_file_name=None,
-                visibility=VisibilityState.TENANT,
-                progress_callback=None):
+    def _prepare_put_request(self,
+                             archive_location,
+                             application_file_name,
+                             visibility,
+                             progress_callback):
         query_params = {'visibility': visibility}
         if application_file_name is not None:
             query_params['application_file_name'] = \
                 urllib.quote(application_file_name)
-
-        uri = '/{self._uri_prefix}/{id}'.format(self=self, id=blueprint_id)
 
         # For a Windows path (e.g. "C:\aaa\bbb.zip") scheme is the
         # drive letter and therefore the 2nd condition is present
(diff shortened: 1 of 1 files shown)


cloudify_rest_client/blueprints.py
```diff
@@ -225,6 +262,48 @@ def upload(self,
...
+        """
+        tempdir = tempfile.mkdtemp()
+        try:
+            tar_path = utils.tar_blueprint(path, tempdir)
+            if not skip_size_limit and os.path.getsize(tar_path) > 30000000:
+                raise Exception('Blueprint folder exceeds 30 MB, '
+                                'move some resources from the blueprint '
+                                'folder to an external location or upload'
+                                ' the blueprint folder as a zip file.')
+            application_file = os.path.basename(path)
```
mabuaisha: The above logic is similar to what we have in upload method, cannot we this logic to share method called `_validte_blueprint_size` for example ?

cloudify_rest_client/blueprints.py
```diff
@@ -225,6 +262,48 @@ def upload(self,
...
+        Validates a blueprint with Cloudify's manager.
+
+        :param path: Main blueprint yaml file path.
+        :param entity_id: Id of the uploaded blueprint.
+        :param visibility: The visibility of the blueprint, can be 'private',
+                           'tenant' or 'global'.
+        :param progress_callback: Progress bar callback method
+        :param skip_size_limit: Indicator whether to check size limit on
+                           blueprint folder
+        :return: Created response.
```
tehasdf: that's a lie ;)

cloudify_rest_client/blueprints.py
```diff
@@ -225,6 +262,48 @@ def upload(self,
...
+
+        Blueprint path should point to the main yaml file of the response
+        to be uploaded. Its containing folder will be packed to an archive
+        and get uploaded to the manager.
+        Validation is basically an upload without storage part being done.
```
tehasdf: s/without storage/without **the** storage/

cloudify_rest_client/blueprints.py
mateumann: You are right; fixed that.

cloudify_rest_client/blueprints.py
mateumann: that's true ;)

cloudify_rest_client/blueprints.py
mateumann: ok

PR #421, opened 2020-03-02
Cy 2366 update cloudify context to include host ip

mabuaisha:


cloudify/utils.py
```diff
@@ -211,7 +211,14 @@ def get_manager_rest_service_host():
...
     """
-    return os.environ[constants.REST_HOST_KEY].split(',')
+    host_ip = None
+    try:
+        # Context could be not available sometimes
+        host_ip = _get_current_context().host_ip
+    except RuntimeError:
+        # Fallback to read from environment variable
+        pass
+    return host_ip or os.environ[constants.REST_HOST_KEY].split(',')
```
tehasdf: this function can be written simpler, just
```python
try:
   return _get_current_context().host_ip
except RuntimeError:
    return os.enviorn[..]
```

cloudify/context.py
```diff
@@ -712,6 +712,10 @@ def workflow_id(self):
         """
         return self._context.get('workflow_id')
 
+    @property
+    def host_ip(self):
```
tehasdf: are you sure `host_ip` is a good name? we already have a `ctx.instance.host_ip` that means the ip of the instance, do you want to use the same name for another thing?

cloudify/workflows/workflow_context.py
```diff
@@ -697,6 +701,7 @@ def _build_cloudify_context(self,
             'execution_id': self.execution_id,
             'workflow_id': self.workflow_id,
             'tenant': self.tenant,
+            'host_ip': self.host_ip,
```
tehasdf: this is not the right place because this also affects local workflows, where it'll just always be none
instead better add it to the remote-specific handler https://github.com/cloudify-cosmo/cloudify-common/blob/8aa6dacf4f26f84a6937a87f320bd81744bed62c/cloudify/workflows/workflow_context.py#L1413

cloudify/utils.py
mabuaisha: @tehasdf You are right but what if the `_get_current_context().host_ip` was `None` for some reason ? this is why I implemented in this way to make sure if the `_get_current_context().host_ip` was `None` to fallback from `enviorn`

cloudify/context.py
mabuaisha: Well, the reason of choosing such name is because `_get_current_context()` would be either `CloudifyContext` instance or `CloudifyWorkflowContex` or any instances extend `_WorkflowContextBase`

cloudify/workflows/workflow_context.py
mabuaisha: You mean to update this as the following 
```
    @property
    def operation_cloudify_context(self):
        return {'local': False,
                'bypass_maintenance': utils.get_is_bypass_maintenance(),
                'rest_token': utils.get_rest_token(),
                'execution_token': utils.get_execution_token(),
                'host_ip': self.workflow_ctx.host_ip}
```

mabuaisha:
@tehasdf I replied to your comments, please let me know what do you think ?

cloudify/context.py
mabuaisha: or we can change it to something else maybe so that it would be common between all type of contex

cloudify/workflows/workflow_context.py
tehasdf: yeah

cloudify/context.py
tehasdf: @mabuaisha yeah after more thought i think i would change it
"host ip", we deal with 100000 hosts, so it's not easy to guess that the "host" means "manager". So probably just "manager_ip" is better
(also it doesn't have to be an ip of course, it can be a hostname, so maybe manager_addr or something, not sure)

cloudify/utils.py
tehasdf: good question what then :)
i'd say that it shouldn't really happen, with the way we have it written right now
maybe the fallback is ok after all

cloudify/context.py
```diff
@@ -557,7 +557,7 @@ def __init__(self, ctx=None):
         self._logger = None
         self._provider_context = None
         self._bootstrap_context = None
-        self._host_ip = None
+        self._host_ip = self._context.get('rest_host')
```
tehasdf: @mabuaisha i see you renamed to `rest_host` but still this `self._host_ip` is used anyway
but this existed beforehand, no?
probably better let's add a new attribute instead of reusing that one, i don't know if it is used by anything else?
i'd say `self._rest_host` for safety, and if `self._host_ip` is not used, then delete it

cloudify/context.py
tehasdf: yeah `rest_host` is a better name

cloudify/context.py
mabuaisha: I totoally agree with you, will use another one instead of this because it could be used by another one

mabuaisha: I cannot see any reference to `_host_id` used so I think we can remove it

PR #422, opened 2020-03-03
CY-2367 Py3 porting: import moves

tehasdf:
Import things in _compat modules, in a py2+3 compatible way

### DIFF
--- a/dsl_parser/import_resolver/abstract_import_resolver.py
+++ b/dsl_parser/import_resolver/abstract_import_resolver.py
@@ -14,8 +14,6 @@
 #  * limitations under the License.
 
 import abc
-import contextlib
-import urllib2
 
 import requests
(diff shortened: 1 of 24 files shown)


dsl_parser/import_resolver/abstract_import_resolver.py
```diff
@@ -57,8 +55,8 @@ def read_import(import_url):
     error_str = 'Import failed: Unable to open import url'
     if import_url.startswith('file:'):
         try:
-            request = urllib2.Request(import_url)
-            with contextlib.closing(urllib2.urlopen(request)) as f:
+            filename = import_url[len('file:'):]
```
geokala: While this should work, it's also a little ugly not stripping the full file:// off the beginning isn't it?

PR #423, opened 2020-03-04
CY-2367 Py3 porting: import moves

tehasdf:
Import things in _compat modules, in a py2+3 compatible way


tehasdf:
duplicate of #422

PR #432, opened 2020-03-16
CY-2460 scheduled tasks snapshot restore

mateumann:
Properly restore scheduled executions from snapshots.  Add `requeue` action for reinserting executions to execution queue after snapshot restoration.

This is a dc9fa569 commit cherry-picked for `5.0.5-ms` branch.


mateumann:
Invalid target branch.  Will repost the PR.

PR #437, opened 2020-03-17
CY-2474 Update node instance to failed state when tasks failed

mabuaisha:


mabuaisha:
> @mabuaisha Do we still want this? If we're not sure, let's close it; if we are sure then let's get it into 6.1.0?

@geokala this is also still need more inputs and feedback from Ofer

geokala:

PR #438, opened 2020-03-22
CY-2315 Add support for selective rollback failed instances

mabuaisha:


cloudify/plugins/workflows.py
```diff
@@ -278,6 +285,139 @@ def validate_inclusions_and_exclusions(include_instances, exclude_instances,
...
+    :param scaled_instances: Represent a list of all scaled in/out instances
+    :param related_instances: Represent a list of related instances
+    generated from scale in/out operation
+    :return: Dict contains both failed instances and related node instances
+    """
+    result = {}
+    failed_instances = set()
+    intact_instances = set()
+    for instance in ctx.node_instances:
+        if instance.state != ready_state:
```
tehasdf: if you reverse this to `if instance.state == ready_state: continue` then the whole block below will be able to be indented less

tehasdf (dismissed):

mabuaisha:
> @mabuaisha Do we still want this one?

@geokala I do not think so since this been suspended for now, so I think for now we can have it here till we decide what to do next since there is a jira opened for this

geokala:

PR #439, opened 2020-03-23
Add `exec_` to compat

tehasdf:
Going to be useful in a few places

### DIFF
--- a/cloudify/_compat.py
+++ b/cloudify/_compat.py
@@ -33,8 +33,13 @@ def reraise(exception_type, value, traceback):
     raise exception_type, value, traceback
 """)
     text_type = unicode
+    exec("""
+def exec_(code, globs):
+    exec code in globs
+""")
 
 else:
+    import builtins
     import queue
     from io import StringIO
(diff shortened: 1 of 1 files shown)


cloudify/_compat.py
```diff
@@ -43,8 +45,10 @@ def reraise(exception_type, value, traceback):
         raise value.with_traceback(traceback)
 
     text_type = str
+    exec_ = getattr(__builtins__, 'exec')
```
tehasdf: gotta do getattr, because `exec_ = __builtins__.exec` or `exec_ = exec` would be a syntaxerror when this is imported under py2

cloudify/_compat.py
```diff
@@ -33,6 +33,10 @@ def reraise(exception_type, value, traceback):
     raise exception_type, value, traceback
 """)
     text_type = unicode
+    exec("""
```
tehasdf: exec-ing `def exec_` to make `exec_`, because otherwise the `exec code in globs` is a syntaxerror when imported under py3

PR #456, opened 2020-04-10
Make pyyaml and networkx be optional installs

tehasdf:
Those aren't required for agents


.circleci/config.yml
```diff
@@ -100,7 +100,7 @@ commands:
             - << parameters.cache_prefix >>-dependencies-{{ checksum "dev-requirements.txt" }}-{{ checksum "test-requirements.txt" }}-{{ checksum "setup.py" }}
       - run: pip install -r dev-requirements.txt --user
       - run: pip install -r test-requirements.txt --user
-      - run: pip install -e . --user
+      - run: pip install -e '.[dispatcher]' --user
```
geokala, line 103: Don't we need to split the tests between those that should need the dispatcher components and those that shouldn't?
Otherwise, we can easily make parts that don't depend on pyyaml/networkx and are used by the agents depend on it at any point without noticing (until the agents break).

tehasdf: currently the split is very simple - cloudify-agent doesn't install `[dispatcher]` in its unittests, and so, those are the tests that shouldn't need the dispatcher

geokala: Makes sense- doesn't fail as early as would be nice, but probably not worth the extra complexity given that it doesn't make sense to use those dependencies in the other components, so marking this as resolved.

PR #460, opened 2020-04-16
CY-2314 Add abort_starting flag for scale_entity (4.5.5-build)

mateumann:
A `scale_entity` workflow accepts a `abort_starting` boolean parameter
(default False) which is passed in context of deployment modification.
Finally this flag shall be used in REST service's resource manager to
clean up stalled/pending dep-modifications before starting new ones.


mateumann:
OBSOLETE

PR #464, opened 2020-04-21
Cy 2488 fix multi network manager

mabuaisha:


cloudify/cluster.py
```diff
@@ -55,7 +55,11 @@ def do_request(self, *args, **kwargs):
             try:
                 return super(ClusterHTTPClient, self).do_request(*args,
                                                                  **kwargs)
-            except (requests.exceptions.ConnectionError):
+            except (requests.exceptions.ConnectionError) as error:
+                self.logger.debug(
+                    'Connection error when trying to connect to '
+                    'manager {0}'.format(error.message)
```
tehasdf, line 61: just `error`

using `.message` is wrong and would fail CI if this was tested

cloudify/context.py
```diff
@@ -714,13 +715,19 @@ def workflow_id(self):
 
     @property
     def rest_host(self):
+        """REST host token"""
```
tehasdf, line 718: fake news? :P

cloudify/workflows/tasks.py
```diff
@@ -540,6 +541,8 @@ def _set_queue_kwargs(self):
                 self._task_tenant = tenant
         self.kwargs['__cloudify_context']['task_queue'] = self._task_queue
         self.kwargs['__cloudify_context']['task_target'] = self._task_target
+        if rest_host:
+            self.kwargs['__cloudify_context']['rest_host'] = rest_host
```
tehasdf, line 545: OK so let's just set this ONLY here,
and not here and in operation_context

so: in the `else` branch below (in the MGMTWORKER_QUEUE) branch, instead of None, return what is now returned in operation_context, ie `self.workflow_ctx.rest_host`

mabuaisha: Sure, make sense will update that

cloudify/context.py
mabuaisha: :P

cloudify/cluster.py
mabuaisha: will update that

PR #470, opened 2020-04-24
CY-2688 Fix lifecycle reinstall_node_instances

mateumann:
Make reinstall_node_instances (part of update workflow) use its own execution
graph not the one generated for uninstall and install steps.

### DIFF
--- a/cloudify/plugins/lifecycle.py
+++ b/cloudify/plugins/lifecycle.py
@@ -22,21 +22,27 @@ from cloudify.workflows.tasks_graph import forkjoin, make_or_get_graph
 from cloudify.workflows import tasks as workflow_tasks
 
 
-def install_node_instances(graph, node_instances, related_nodes=None):
+def install_node_instances(graph,
+                           node_instances,
+                           related_nodes=None,
+                           name_prefix=''):
     processor = LifecycleProcessor(graph=graph,
                                    node_instances=node_instances,
-                                   related_nodes=related_nodes)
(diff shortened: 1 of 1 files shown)


cloudify/plugins/lifecycle.py
```diff
@@ -45,15 +51,16 @@ def reinstall_node_instances(graph,
                              ignore_failure,
                              related_nodes=None):
     uninstall_node_instances(graph, node_instances, ignore_failure,
-                             related_nodes)
+                             related_nodes, name_prefix='reinstall')
```
tehasdf, line 54: maybe `reinstall-` with the dash so that the name is then `reinstall-install`, not `reinstallinsall`

mateumann: I thought about it and maybe single `re` would do?

BTW there are entries with `name` like `unlinkuninstall` or `establishinstall` in `tasks_graphs`.  Maybe we should/could do something about it as well.

tehasdf: go ahead :D 
https://github.com/cloudify-cosmo/cloudify-common/blob/1d9e1e0301e3bf5378cdbc1b002a67ad9771e319/cloudify/plugins/lifecycle.py#L67
and here
https://github.com/cloudify-cosmo/cloudify-common/blob/1d9e1e0301e3bf5378cdbc1b002a67ad9771e319/cloudify/plugins/lifecycle.py#L79

PR #493, opened 2020-06-25
CY-2822 Fix blueprint upload folder size check

glukhman:


cloudify_rest_client/utils.py
```diff
@@ -60,3 +60,17 @@ def is_kerberos_env():
     if os.path.exists('/etc/krb5.conf') and find_executable('klist'):
         return True
     return False
+
+
+def get_folder_size_and_files(path):
```
geokala, line 65: Can you add a unittest for this in case someone does something clever with it later?

PR #495, opened 2020-07-01
increased number of allowed files

AdarShaked:


geokala:
Just discussed and this will be made configurable soon, but we are increasing the base value for the moment.

PR #497, opened 2020-07-07
CY-3015 Install plugins on first use

tehasdf:
When running an operation, check if the plugin is installed,
and if it's not - install it then.

This is as opposed to the previous approach of always installing
plugins up-front.

Also, pass through the plugin source to operations.

### DIFF
--- a/cloudify/dispatch.py
+++ b/cloudify/dispatch.py
@@ -50,6 +50,12 @@ from cloudify.error_handling import (
     serialize_known_exception,
     deserialize_known_exception
 )
+
+try:
+    from cloudify_agent.api.plugins import installer as plugin_installer
+except ImportError:
+    plugin_installer = None
+
 try:
     from cloudify.workflows import api
     from cloudify.workflows import workflow_context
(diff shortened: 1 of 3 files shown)


cloudify/utils.py
```diff
@@ -543,9 +543,14 @@ def _get_formatted_version(version):
...
         # get all plugin dirs
-        subdirs = next(os.walk(plugins_dir))[1]
+        try:
+            subdirs = next(os.walk(plugins_dir))[1]
+        except StopIteration:
+            return
         # filter by package name
         package_dirs = [dir for dir in subdirs if dir.startswith(package_name)]
+        if not package_dirs:
+            return
```
tehasdf, line 553: plugins are now not required to be there, in case they're not installed yet, so cope with that

cloudify/dispatch.py
```diff
@@ -50,6 +50,12 @@
     serialize_known_exception,
     deserialize_known_exception
 )
+
+try:
+    from cloudify_agent.api.plugins import installer as plugin_installer
+except ImportError:
```
tehasdf, line 56: keep this optional so that operations can be run without agent installed (tests/cfy local)

PR #501, opened 2020-07-09
Don't str unnecessarily, to avoid unicode errors

geokala:


cloudify/utils.py
```diff
@@ -360,7 +360,7 @@ def exception_to_error_cause(exception, tb):
     etype = type(exception)
     traceback.print_exception(etype, exception, tb, file=error)
     return {
-        'message': str(exception),
+        'message': exception,
```
tehasdf, line 363:
```suggestion
        'message': u'{0}'.format(exception),
```
for a 2+3 compatible turning the exception into unicode
(as opposed to the previous str()-ing, which wasa turning it into unicode or bytes)

geokala: Updating to use that approach

PR #503, opened 2020-07-13
Only encode env vars on py2

geokala:
On py3 it raises exceptions.


tehasdf:
this is because py3 accepts unicode there, and on windows, it absolutely demands unicode. On py2, it absolutely demands bytes on both windows and linux.

PR #504, opened 2020-07-13
When checking for plugin dir, check for the marker file as well

tehasdf:
Instead of just checking .isdir, also look for the plugin.id file.
Otherwise it will be just a directory and not a directory containing
a plugin.

In practice, it will be a plugin that is not fully installed yet,
so we should go to the installer, which will do the locking-and-waiting
dance.

### DIFF
--- a/cloudify/utils.py
+++ b/cloudify/utils.py
@@ -561,6 +561,12 @@ class Internal(object):
         return newest
 
     @staticmethod
+    def _is_plugin_dir(path):
+        """Is the given path a directory containing a plugin?"""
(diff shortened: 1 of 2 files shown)


cloudify/utils.py
```diff
@@ -560,6 +560,12 @@ def _get_package_version(plugins_dir, package_name):
             return
         return newest
 
+    @staticmethod
```
tehasdf, line 563: i hate `@staticmethod` but eh, i'll just do it here for consistency with all the other methods

PR #506, opened 2020-07-20
CY-3111 Replace certificates on agents

YoniItzhak:


cloudify_rest_client/utils.py
```diff
@@ -74,3 +74,10 @@ def get_folder_size_and_files(path):
             size += sub_size
             files += sub_files
     return size, files
+
+
+def get_file_content(file_path):
+    if os.path.exists(file_path):
+        with open(file_path) as fd:
+            return fd.read()
+    return None
```
geokala, line 83: I'm torn between whether this should return None or just try to open and raise an error- any reason you chose this approach?

Either way, approved.

YoniItzhak: I figured this approach is safer :)

PR #507, opened 2020-07-21
CY-3094 REST-client calls for forcing plugin installation

tehasdf:


cloudify/dispatch.py
```diff
@@ -98,7 +98,9 @@
     'script',
     'cfy_extensions',
     'default_workflows',
-    'worker_installer'
+    'worker_installer',
+    'cloudify_system_workflows'
+
```
geokala, line 103: Well, I do like trailing something- but generally comma, not extra newline.

cloudify/context.py
```diff
@@ -352,6 +352,11 @@ def id(self):
         """The node instance id."""
         return self._context.get('node_id')
 
+    @property
```
geokala, line 355: Any reason not to just make self._node be self.node?

cloudify/models_states.py
```diff
@@ -93,3 +93,13 @@ class AgentState(object):
     STATES = [CREATING, CREATED, CONFIGURING, CONFIGURED, STARTING, STARTED,
               STOPPING, STOPPED, DELETING, DELETED, RESTARTING, RESTARTED,
               UPGRADING, UPGRADED, FAILED, NONRESPONSIVE, RESTORED]
+
+
+class PluginInstallationState(object):
+    PENDING = 'pending'
```
geokala, line 99: Would it makes sense to have this be pending-install, to mirror pending-uninstall?

cloudify/context.py
tehasdf: consistency with all the other ones

tehasdf: and i guess self._node is a restclient node object, only used internally in this, and this is a NodeContext

PR #510, opened 2020-07-28
CY-3162 Fix maintenance mode bypass on bootstrap context

glukhman:


geokala:
Agreed, let's migrate away from env vars for this- good reasoning.

PR #512, opened 2020-07-28
Refactor utils.plugin_prefix

tehasdf:
This makes getting the plugin directory a bit simpler.

Plugins are now installed in, for managed plugins:
`/opt/mgmtworker/env/plugins/<tenant>/<plugin name>/<version>`
and for source plugins:
`/opt/mgmtworker/env/source_plugins/<tenant>/<plugin name>/<version>`

Also, visibility isn't needed, because plugins are installed for
the _using_ tenant anyway. And if visibility isn't global, then
the plugin can't be downloaded or used anyway.

### DIFF
--- a/cloudify/dispatch.py
+++ b/cloudify/dispatch.py
@@ -29,7 +29,6 @@ import traceback
 from contextlib import contextmanager
 
 from cloudify_rest_client.executions import Execution
-from cloudify_rest_client.constants import VisibilityState
 from cloudify_rest_client.exceptions import (
     InvalidExecutionUpdateStatus,
     CloudifyClientError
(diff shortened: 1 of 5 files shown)


cloudify/dispatch.py
```diff
@@ -288,15 +287,17 @@ def dispatch_to_subprocess(self):
                 os.environ[constants.BYPASS_MAINTENANCE] = 'True'
             env = self._build_subprocess_env()
 
-            plugin_dir = self._extract_plugin_dir()
-            if not plugin_dir and self._should_install_plugin():
-                self._install_plugin()
+            if not self._should_install_plugin():
```
tehasdf, line 290: todo flip this condition, as soon as i figure out a better name than `should_install_plugin`

cloudify/tests/test_context.py
```diff
@@ -299,19 +299,17 @@ def test_prefix_from_wagon(self):
...
+            self.deployment_id,
+            self.plugin_pacakge_name,
+            self.plugin_pacakge_version)
         os.makedirs(expected_prefix)
         with open(os.path.join(expected_prefix, 'plugin.id'), 'w') as f:
             f.write('plugin id')
         with patch('sys.prefix', self.test_prefix):
             self.assertEqual(self.ctx.plugin.prefix, expected_prefix)
 
-    def test_fallback_prefix(self):
```
tehasdf, line 312: the fallback is gone and so the test goes as well

cloudify/tests/test_local_workflows.py
```diff
@@ -755,7 +755,6 @@ def ctx_properties(ctx, **_):
             self.assertEqual(PLUGIN_PACKAGE_NAME, ctx.plugin.package_name)
             self.assertEqual(PLUGIN_PACKAGE_VERSION,
                              ctx.plugin.package_version)
-            self.assertEqual(sys.prefix, ctx.plugin.prefix)
```
tehasdf, line 758: fallback to the main venv is absolutely disallowed

cloudify/tests/test_context.py
```diff
@@ -288,8 +288,8 @@ def test_prefix_from_wagon(self):
             self.test_prefix,
             'plugins',
             self.tenant_name,
-            '{0}-{1}'.format(self.plugin_pacakge_name,
-                             self.plugin_pacakge_version))
+            self.plugin_pacakge_name,
```
geokala, line 291: pacakge?

PR #515, opened 2020-07-30
contexts: make bypass_maintenance a public attribute

tehasdf:
Make this a real attribute, so that 'external' methods
(the one in utils.py) use a public attribute on the context,
instead of touching the internal ._context

This matters because quasi-contexts such as the CancelCloudifyContext
(which is used for kill-canceling) or the PluginInstallContext
(plugin force-installing) declare some attributes, and ._context
is not one of them. I want to make those quasi-contexts declare
.bypass_maintenance, I don't want to make them declare an
'internal' .context

### DIFF
--- a/cloudify/context.py
+++ b/cloudify/context.py
@@ -734,6 +734,11 @@ class CloudifyContext(CommonContext):
         return self._rest_ssl_cert
 
     @property
+    def bypass_maintenance(self):
+        """If true, all requests sent bypass maintenance mode."""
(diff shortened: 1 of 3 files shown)


tehasdf:
in here: https://github.com/cloudify-cosmo/cloudify-manager/blob/8e2cdd0d1a9ac7c3542edbe58aaf4569ef217afe/mgmtworker/mgmtworker/worker.py#L134-L138
it's going to be nicer to declare `.bypass_maintenance` than `.context`

PR #527, opened 2020-08-10
CY-3138 Get rid of distutils import

glukhman:


cloudify_rest_client/utils.py
```diff
@@ -81,3 +81,27 @@ def get_file_content(file_path):
...
     return None
+
+
+def find_executable(executable, path=None):
+    """Tries to find 'executable' in the directories listed in 'path'.
+    A string listing directories separated by 'os.pathsep'; defaults to
+    os.environ['PATH'].  Returns the complete filename or None if not found.
+    ~~NOTE:~~
+    Copied verbatim from distutils.spawn, since we don't always ship distutils.
+    """
```
tehasdf, line 92:
```suggestion
# Copied verbatim from distutils.spawn, since we don't always ship distutils.
def find_executable(executable, path=None):
    """Tries to find 'executable' in the directories listed in 'path'.

    A string listing directories separated by 'os.pathsep'; defaults to
    os.environ['PATH'].  Returns the complete filename or None if not found.
    """
```
I think like this is better, because the "copied from" note is then outside the copied code, and not somewhere in the middle of it

(also in this suggestion I'm re-adding the newline; note that "1 summary line, newline, long description" is the standard format for docstrings per PEP257)

PR #528, opened 2020-08-11
CY-3316 Fix the prometheus case

YoniItzhak:


YoniItzhak:
The CLI used the function `change_cluster_cert` in order to replace the client certificate at the end of the certificates' replacement. I created a function for this since the `_client` attribute is supposed to be protected. But there might be a better solution.

PR #530, opened 2020-08-12
CY-3272 Add username to mgmtworker logs

YoniItzhak:


cloudify/amqp_client.py
```diff
@@ -476,8 +476,15 @@ def _log_message(self, message):
         level = message.get('level', 'info')
         log_func = getattr(self.logger, level, self.logger.info)
         exec_id = message.get('context', {}).get('execution_id')
+        execution_creator_username = message.get('context', {}).get(
```
geokala, line 479: This might also benefit from the logging changes we're discussing in the other PR.

PR #540, opened 2020-08-26
CY-3237 Amqp client: reconnect on channel close

tehasdf:
The "queue doesn't exist" would be a programming error and it
doesn't happen in normal operation. What does happen, is getting
a channel close on rabbitmq failovers, and we should just resume
operation then. By reconnecting.


geokala:
This appears to have been implemented in different PRs judging by the closed ticket. Closing.

PR #542, opened 2020-09-02
CY-3419 Restore operations belonging to finished subgraphs

tehasdf:
We have the logic that if a task is already finished, we don't restore it to the graph.

However, if the task was failed, the subgraph would've been failed as well. And when we
resumed that failed task with reset-operations, the subgraph wouldn't be part of
the graph.
This makes it so that a subgraph to which a resumed task belongs, is also going to
be resumed. Its state is reset to STARTED.

Also added a few unittests.

Note that commit 2 implements the idea, but commit 3 refactors it, so it probably
doesn't make sense to read too much into commit 2 on its own.

### DIFF
--- a/cloudify/tests/test_tasks_graph.py
+++ b/cloudify/tests/test_tasks_graph.py
@@ -18,6 +18,8 @@ import time
 import testtools
 from contextlib import contextmanager
 
+from cloudify_rest_client.operations import Operation
+
 from cloudify.workflows import api
 from cloudify.workflows import tasks
 from cloudify.workflows.tasks_graph import TaskDependencyGraph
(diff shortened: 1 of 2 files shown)


cloudify/tests/test_tasks_graph.py
```diff
@@ -173,3 +175,98 @@ def test_cancel(self):
...
+        """Restoring an empty list of operations results in an empty graph"""
+        graph = self._restore_graph([])
+        operations = list(graph.tasks_iter())
+        assert operations == []
+
+    def test_restore_single(self):
+        """A single operation is restored into the graph"""
+        graph = self._restore_graph([self._remote_task()])
+        operations = list(graph.tasks_iter())
+        assert len(operations)
```
geokala, line 231: Shouldn't you check this == 1?

geokala (approved):
Looks good, aside from possibly wanting a more specific assert as mentioned in my previous comment (which you fixed while I was looking at it, so LGTM).

PR #544, opened 2020-09-04
CY-3446 cfy plugin update enhancement

mateumann:


cloudify_rest_client/plugins_update.py
```diff
@@ -106,21 +106,29 @@ def list(self, _include=None, sort=None, is_descending=False, **kwargs):
...
+    def update_plugins(self, blueprint_id, force=False,
+                       plugin_name=None, minor=False, minor_except=None):
         """
         Updates the plugins in all the deployments that use the given
         blueprint.
 
         :param blueprint_id: blueprint ID to perform the update with.
         :param force: if to forcefully update when other non-active plugins
          updates exists associated with this blueprint [deprecated].
+        :param plugin_name: list of plugin names to update (only those).
```
geokala, line 118: Nitpick: Shouldn't it be called plugin_names if it's expected to be a list?

cloudify_rest_client/plugins_update.py
```diff
@@ -106,21 +106,29 @@ def list(self, _include=None, sort=None, is_descending=False, **kwargs):
...
+        :param minor: update all selected plugins to the latest minor release.
+        :param minor_except: list of plugin names; all plugins will be updated
```
geokala, line 120: This looks like a parameter to exclude specific minor versions. Perhaps it should instead be called to_latest?

cloudify_rest_client/plugins_update.py
```diff
@@ -106,21 +106,29 @@ def list(self, _include=None, sort=None, is_descending=False, **kwargs):
...
+                       plugin_name=None, minor=False, minor_except=None):
         """
         Updates the plugins in all the deployments that use the given
         blueprint.
 
         :param blueprint_id: blueprint ID to perform the update with.
         :param force: if to forcefully update when other non-active plugins
          updates exists associated with this blueprint [deprecated].
+        :param plugin_name: list of plugin names to update (only those).
+        :param minor: update all selected plugins to the latest minor release.
```
geokala, line 119: Is this a flag or is this another list of plugin names? If the latter, maybe it would be better being something like update_only_minor?

mateumann: it's a flag (or was ;) )

PR #547, opened 2020-09-08
REST Client: handle 204 responses

tehasdf:
204 means no content, so no reason to try and json-decode it.
There's no content. Don't even try.


geokala (approved):
I think the function is misnamed but that's not on you, so approved.

PR #548, opened 2020-09-10
CY-3456 Smart resume for the install workflow

tehasdf:
When the install workflow is resumed, if we are to re-send
the create operation, send a delete first. If we are to re-send
the stop operation, send stop first.

### DIFF
--- a/cloudify/plugins/lifecycle.py
+++ b/cloudify/plugins/lifecycle.py
@@ -154,7 +154,7 @@ class LifecycleProcessor(object):
                 'resume_cleanup_{0}'.format(instance.id))
             sequence = uninstall_subgraph.sequence()
             sequence.add(*tasks)
-            ignore_subgraph_on_task_failure(uninstall_subgraph, instance)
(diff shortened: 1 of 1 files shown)


cloudify/plugins/lifecycle.py
```diff
@@ -119,6 +121,41 @@ def uninstall(self):
...
+            if instance.state in ['starting'] and _would_resend(
+                    install_subgraph, 'cloudify.interfaces.lifecycle.start'):
+                tasks += _pre_resume_stop(instance)
+            if instance.state in ['creating'] and _would_resend(
+                    install_subgraph, 'cloudify.interfaces.lifecycle.create'):
+                tasks += _pre_resume_uninstall(instance)
+            if not tasks:
+                continue
+            tasks = [
+                instance.send_event('Uninstalling prior to resuming install')
```
geokala, line 147: This message isn't quite accurate.

geokala (approved):

PR #551, opened 2020-09-15
CY-3456 NOP tasks: pass through the id, keep them in the graph

tehasdf:
In the various workflow graphs, there are still NOP tasks
used for structuring the graph and keeping dependencies.

During graph restore, we key the dicts by task id, and so the
task id must be passed through and set. Hence making sure the
kwargs are passed.

Since the NOP tasks are used for structuring, they must go into
the restored graph as well. Hence storing them as pending after all.
They will always go into the graph, and possibly be executed
immediately after restoring. But they will always be there and will
influence the ordering of other operations as they should.


tehasdf:
The integration test for this is cloudify-cosmo/cloudify-manager#2338

geokala (approved):

PR #554, opened 2020-09-21
CY-3420 Support DELETE calls returning 204

tehasdf:


cloudify_rest_client/events.py
```diff
@@ -100,7 +100,8 @@ def delete(self, deployment_id, include_logs=False, message=None,
                                     deployment_id=deployment_id,
                                     **kwargs)
 
-        response = self.api.delete(uri, params=params)
+        response = self.api.delete(uri, params=params,
+                                   expected_status_code=200)
```
geokala, line 104: Not 204?

cloudify_rest_client/executions.py
```diff
@@ -338,5 +338,6 @@ def delete(self, to_datetime=None, keep_last=None, **kwargs):
             data['keep_last'] = keep_last
         response = self.api.delete('/{self._uri_prefix}'.format(self=self),
                                    data=data,
-                                   params=kwargs)
+                                   params=kwargs,
+                                   expected_status_code=200)
```
geokala, line 342: Not 204?

cloudify_rest_client/events.py
tehasdf: in events and executions, we do return the deleted count. Which is, as you know, allowed by the rfc.

geokala (dismissed):

cloudify_rest_client/events.py
tehasdf: CY-3535 is about updating the rest docs to reflect all these changes

geokala (dismissed):

geokala (approved):
Any more for any more?

PR #560, opened 2020-09-24
Add a way for operations to tell the agent to stop

tehasdf:


cloudify/amqp_client.py
```diff
@@ -413,14 +414,20 @@ def _process_message(self, channel, properties, full_task, delivery_tag):
...
                     'routing_key': properties.reply_to,
                     'properties': pika.BasicProperties(
                         correlation_id=properties.correlation_id),
-                    'body': json.dumps(result)
+                    'body': body
                 })
-
+        if result is STOP_AGENT:
+            # the operation asked us to exit, so drop everything and exit
+            os._exit(0)
```
tehasdf, line 430: when I say "drop everything", I mean it. Hence the very forceful `_exit`.
The previous approach was to `kill -9` the daemon from outside so this is just a bit nicer.

(regular `sys.exit` would wait for all the threads to finish, which doesn't really ever happen)

tehasdf:
ignore the new jenkins CI, it fails due to unrelated CY-3570

geokala (approved):

PR #561, opened 2020-09-25
Make delete calls work with pre 5.1 managers

geokala:
Otherwise system tests all break.


geokala:
Closing this, as we'll go with @tehasdf 's approach.

PR #563, opened 2020-09-28
CY-3570 Get rid of the fileserver in unittests

tehasdf:
Instead of starting a fileserver and using it in tests
for get_resource, mock out requests.get.

This makes the tests actual unittests, then.

Directly, this is because the port for the fileserver was hardcoded
and not changing, which meant we couldn't run multiple runs
of these tests at the same time, which led to random failures
on the new jenkins.


cloudify/tests/test_context.py
```diff
@@ -87,58 +77,77 @@ def redirect_log_to_stdout(logger):
...
     def test_get_resource(self, _):
-        resource = self.context.get_resource(
-            resource_path='for_test_bp_resource.txt')
+        with patch('requests.get',
+                   return_value=Mock(ok=True, content=b'Hello from test')):
+            resource = self.context.get_resource(
+                resource_path='for_test_bp_resource.txt')
         self.assertEquals(resource, b'Hello from test')
 
-    def test_get_deployment_resource_priority_over_blueprint_resource(self):
```
tehasdf, line 96: this tests the ordering of tried urls, which is already tested by `test_download_resource_tried_urls`

cloudify/tests/test_context.py
```diff
@@ -87,58 +77,77 @@ def redirect_log_to_stdout(logger):
...
-        deployment_context_mock = MagicMock()
-        deployment_context_mock.id = 'dep1'
-        self.context.deployment = deployment_context_mock
-        resource = self.context.get_resource(resource_path='for_test.txt')
-        self.assertEquals(resource, b'belongs to dep1')
-
-    def test_get_deployment_resource_no_blueprint_resource(self):
```
tehasdf, line 103: this tests that a deployment resource is tried, which is already tested by `test_download_resource_tried_urls`

tehasdf (commented):
removed 2 tests, and commented why

geokala (approved):

PR #572, opened 2020-10-05
CY-3619 Plugin state update: fetch plugin if missing

tehasdf:
If we don't have the plugin details, attempt to fetch them.
This is similar, and uses the same method as, installing the plugin.

This is because sometimes (eg. in the uninstall in plug-update)
we only get package-name and package-version but not the id.


tehasdf:
this fixes the asserts added to the integration tests in cloudify-cosmo/cloudify-manager#2374

mateumann (approved):

PR #576, opened 2020-10-05
get_resource: try all possible paths

tehasdf:
get_resource is supposed to try 3 paths:
  - deployment-local  (for all managers)
  - blueprint-local  (for all managers)
  - manager-global  (for all managers)
..for up to `3 * len(managers)` calls in a cluster.

It used to, if one of the managers was offline, throw a NonRecoverable
on possibly the deployment-local stage.

This meant that if one manager was offline, it was impossible to
use get_resource to get blueprint-local or manager-global
resources.

Instead, make sure that get_resource never throws nonrecoverable,
always throws httpexception if it doesn't found, but most importantly-
always tries all managers, even if some managers are offline.
(I'll keep that `if` as 404 for now, I don't think that needs to
be made more lenient)

### DIFF
--- a/cloudify/manager.py
+++ b/cloudify/manager.py
@@ -356,6 +356,8 @@ def get_resource(blueprint_id, deployment_id, tenant_name, resource_path):
             blueprint_id, deployment_id, tenant_name, resource_path):
         try:
             return get_resource_from_manager(path)
+        except NonRecoverableError:
+            tried_paths.append(path)
         except HttpException as e:
             if e.code != 404:
                 raise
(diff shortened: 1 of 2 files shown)


cloudify/manager.py
```diff
@@ -356,6 +356,8 @@ def get_resource(blueprint_id, deployment_id, tenant_name, resource_path):
             blueprint_id, deployment_id, tenant_name, resource_path):
         try:
             return get_resource_from_manager(path)
+        except NonRecoverableError:
```
tehasdf, line 359: i realize `except NonRecoverable` is ugly, because it's supposed to be "non recoverable" :) but this is the only backwards-compatible way for both functions (get_resource and get_resource_from_manager), and both are kindasorta public...

mateumann (approved):

PR #579, opened 2020-10-13
Bump PyYaml to 4.2b4

glukhman:


geokala (approved):

glukhman:
The reason we bump to 4.2b4 and not 5.X is that the next version of PyYaml, 5.1, has dropped support for Python 2.6, which we still need

PR #584, opened 2020-10-19
CY-3689 Add monitoring module for handling Prometheus

mateumann:
Two methods added: `manager_added` and `manager_removed` which handle manager
being added/removed from the database.  Currently that means only re-rendering
`/etc/prometheus/targets/other_managers.yml`.


mateumann:
`monitoring.py` has been moved to the `cloudify-manager/mgmtworker/mgmtworker/`.

PR #586, opened 2020-10-27
RD-326 Blueprint input from datatypes handling

mateumann:
For some blueprints default values for the inputs are stored inside
data_types.  Before this patch those were not taken into consideration
while setting plan inputs.

The new approach is parsing an empty value in case no input or input's
default value were provided.

### DIFF
--- a/dsl_parser/tasks.py
+++ b/dsl_parser/tasks.py
@@ -60,12 +60,34 @@ def _set_plan_inputs(plan, inputs=None):
                     exceptions.ERROR_INVALID_CHARS,
                     'Illegal characters in input: {0}. '
                     'Only valid ascii chars are supported.'.format(input_name))
+        elif DEFAULT in input_def and input_def[DEFAULT] is not None:
+            inputs[input_name] = input_def[DEFAULT]
         else:
-            if DEFAULT in input_def and input_def[DEFAULT] is not None:
-                inputs[input_name] = input_def[DEFAULT]
-            else:
+            # Try to get some defaults from the data_type maybe or in another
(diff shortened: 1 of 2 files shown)


dsl_parser/tasks.py
```diff
@@ -60,12 +60,34 @@ def _set_plan_inputs(plan, inputs=None):
...
                     'Only valid ascii chars are supported.'.format(input_name))
+        elif DEFAULT in input_def and input_def[DEFAULT] is not None:
+            inputs[input_name] = input_def[DEFAULT]
         else:
-            if DEFAULT in input_def and input_def[DEFAULT] is not None:
-                inputs[input_name] = input_def[DEFAULT]
-            else:
+            # Try to get some defaults from the data_type maybe or in another
+            # words just try to parse the value before validating it.
+            try:
```
geokala, line 68: Does this mean that if someone provides a default value we don't check the type for it?
Maybe put a couple of unit(ish?) tests in to confirm the different behaviours of this?

mateumann: No, we still do validate/verify inputs (see lines (92++).  This addition is only to make sure we try to retrieve/parse value from `data_type` first (and not report input as missing).

geokala: Excellent,

geokala (commented):
Perhaps we should also have an extra test or two covering behaviour if someone, e.g. uses this blueprint but sets test_config to a correct value and to the value "beep" (or something else that isn't the correct custom type)- to catch any errors in later updates to this code.

geokala (approved):

PR #589, opened 2020-11-02
RD-208 Replace tasks polling with callbacks

tehasdf:
Rewrite the TasksGraph, and the WorkflowTasks themselves, to use callbacks
instead of polling task state.

This makes the graph execution faster and less resource-intensive, as well as
lower latency; also, the code is more readable (I hope!) as there's no more
things like multi-line conditions in `_executable_tasks`, no networkx graphs,
and less moving parts in the tasks themselves and the _TaskDispatcher.

Also some changes to the tests.

See each commit message for details.

BTW, @geokala, this WOULD be net -200 lines, it's only +60 because of the types.yaml! :P

### DIFF
--- a/cloudify/dispatch.py
+++ b/cloudify/dispatch.py
@@ -673,8 +673,7 @@ class WorkflowHandler(TaskHandler):
             self.ctx.internal.start_local_tasks_processing()
             result = self.func(*self.args, **self.kwargs)
             if not self.ctx.internal.graph_mode:
-                tasks = list(self.ctx.internal.task_graph.tasks_iter())
-                for workflow_task in tasks:
+                for workflow_task in self.ctx.internal.task_graph.tasks:
                     workflow_task.async_result.get()
             return result
         finally:
--- a/cloudify/dispatch.py
+++ b/cloudify/dispatch.py
@@ -614,13 +614,13 @@ class WorkflowHandler(TaskHandler):
                     # deprecated api.EXECUTION_CANCELLED_RESULT as result).
                     # parent thread then goes back to polling for messages from
                     # child thread or possibly 'force-cancelling' requests
-                    api.cancel_request = True
+                    api.set_cancel_request()
 
                 if execution.status == Execution.KILL_CANCELLING:
                     # if a custom workflow function must attempt some cleanup,
                     # it might attempt to catch SIGTERM, and confirm using this
                     # flag that it is being kill-cancelled
-                    api.kill_request = True
+                    api.set_kill_request()
 
                 if execution.status in [
                         Execution.FORCE_CANCELLING,
--- a/cloudify/workflows/tasks.py
+++ b/cloudify/workflows/tasks.py
@@ -793,6 +793,10 @@ class WorkflowTaskResult(object):
             self.result = api.ExecutionCancelled()
             raise self.result
 
+        ctx = self.task.workflow_context
+        if not ctx.internal.graph_mode:
+            ctx.internal.task_graph.remove_task(self.task)
+
         if self.task.get_state() in (TASK_FAILED, TASK_RESCHEDULED):
             handler_result = self.task.handle_task_terminated()
             if handler_result.retried_task and retry_on_failure:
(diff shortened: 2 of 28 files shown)


tehasdf:
not removing networkx because the dsl-parser still uses it :(

cloudify/workflows/workflow_api.py
```diff
@@ -19,6 +19,26 @@
...
+    cancel_request = True
+    for f in cancel_callbacks:
+        f()
+
+
+def set_kill_request():
+    global cancel_request, kill_request
+    cancel_request = True
+    kill_request = True
+    for f in kill_request:
```
geokala, line 37: Shouldn't this be 'for f in kill_callbacks'?

cloudify/dispatch.py
```diff
@@ -673,8 +673,7 @@ def _execute_workflow_function(self):
             self.ctx.internal.start_local_tasks_processing()
             result = self.func(*self.args, **self.kwargs)
             if not self.ctx.internal.graph_mode:
-                tasks = list(self.ctx.internal.task_graph.tasks_iter())
-                for workflow_task in tasks:
+                for workflow_task in self.ctx.internal.task_graph.tasks:
```
geokala, line 676: Why was this tasks_iter(), and why doesn't it need to be any more?

tehasdf: this looked at every task, for reasons I describe in the `task .get: remove from graph` commit; it still does, I'm just changing the method

cloudify/workflows/tasks_graph.py
```diff
@@ -466,18 +454,33 @@ def add_dependency(self, src_task, dst_task):
         self.graph.add_dependency(src_task, dst_task)
 
     def apply_async(self):
+        super(SubgraphTask, self).apply_async()
         if not self.tasks:
             self.set_state(tasks.TASK_SUCCEEDED)
         else:
+            for task_id, task in self.tasks.items():
+                self.graph.remove_dependency(task, self)
```
geokala, line 462: I may be misreading this, but it looks a bit like we claim the dependency is resolved the moment we start a task rather than on finishing it?

tehasdf: yeah. As I noted in one of the commits, it used to be that the "task runs after its graph" was encoded somewhere in the `_executable_task` stack of `if`s. Now it's directed by the dependency that tasks have on their subgraphs. So when a subgraph starts, we remove that dependency, so that the tasks can run.

This is as opposed to finishing the subgraph right there, because we do want the subgraph to only finish when its tasks finish.

An alternative would be additional `if`s inside of the graph itself. This approach turned out to be cleaner.

Also added a comment to the code saying as much

geokala: But doesn't this mean that something that depends on task x will (be able to) start running as soon as task x starts, without waiting for it to finish?

That could presumably result in something failing or needing to retry due to the actions task x should take not yet being complete?

cloudify/workflows/tasks.py
```diff
@@ -32,15 +32,14 @@
     TASK_SENDING,
     TASK_SENT,
     TASK_STARTED,
-    TASK_RESPONSE_SENT,
     TASK_RESCHEDULED,
     TASK_SUCCEEDED,
     TASK_FAILED,
-    TERMINATED_STATES
+    TERMINATED_STATES,
```
geokala, line 38: It's almost like you knew I was going to be reading this...

cloudify/workflows/tasks_graph.py
tehasdf: you're reading the implementation of a SubgraphTask here, so yes, but only if task x is a subgraph.

Of course, tasks inside of the subgraph can and will have also other dependencies that they will have to wait for: eg. in a subgraph S containing a sequence of tasks [A, B, C]:
  - task A has dependencies on S,
  - task B has dependencies on S, A
  - task C has dependencies on S, B
so starting the subgraph resolves that S dependency, allowing A to run, but B will still wait for A (but not for S anymore)

cloudify/workflows/tasks.py
tehasdf: should've made it comma-first

cloudify/workflows/tasks_graph.py
geokala: Ah, ok, that makes sense. Marking as resolved then.

geokala (approved):

PR #591, opened 2020-11-09
RD-406 Support for storing .on_failure/.on_success handlers

tehasdf:
Commit 1 introduces the infra for serializing/deserializing them, and commit 2
implements that for builtin cases


tehasdf:
I decided not to use pickle after all, because this way feels more predictable

geokala (approved):

PR #593, opened 2020-11-16
WorkflowTaskResult.get: support cancel when task is already done

tehasdf:
When the task gets finished, but we received a cancel request in
the meantime, just return the finish value.

Only throw ExecutionCancelled if we were cancelled and the task
has no return value (yet).

In other words, if the task finishes at the same time as the execution
gets cancelled, the task finish value gets priority. Because well,
it did finish.


tehasdf:
fixes `integration_tests/tests/agentless_tests/test_executions.py -k test_legacy_cancel_execution`

glukhman (approved):

PR #595, opened 2020-11-16
with_execute_after: push workflow context

tehasdf:
the original .apply_async() is called in workflow context, so
when it is delayed and called in a different thread - the timer thread-
it needs to be in wokrflow context as well.
We have the workflow context on the task, so let's just push it.


tehasdf:
fixes `integration_tests/tests/agentless_tests/test_task_retries.py -k test_infinite_retries`
(otherwise that test throws a not-in-context exception into the deployment log)

glukhman (approved):

PR #596, opened 2020-11-16
task duplicate: also duplicate success/failure handlers

tehasdf:
The task is being duplicated, so we must copy these as well.


tehasdf:
fixes `integration_tests/tests/agentless_tests/test_task_retries.py -k  test_retries_ignore_total`
otherwise the get_host_state task is only retried _once_, because the retry doesn't have the .on_success anymore

glukhman (approved):

PR #597, opened 2020-11-19
For SSL certificate verification, use a more specialized error message

isaac-s:


cloudify_rest_client/client.py
```diff
@@ -279,6 +279,17 @@ def do_request(self,
...
+            # Special handling: SSL Verification Error.
+            # We'd have liked to use `__context__` but this isn't supported in
+            # Py26, so as long as we support Py26, we need to go about this
+            # awkwardly.
+            if len(e.args) > 0 and 'CERTIFICATE_VERIFY_FAILED' in str(
+                    e.args[0]):
+                raise requests.exceptions.SSLError(
+                    'Certificate verification failed; please ensure that the '
+                    'certificate presented by Cloudify Manager is trusted by '
+                    'your operating system (underlying reason: {0})'.format(
```
geokala, line 291: I'm not sure the 'by your operating system' bit of this is necessary- e.g. if it's in the CLI then the OS cert store isn't used.

Otherwise looks as reasonable as it can be I think.

isaac-s: Amended

geokala (approved):

PR #598, opened 2020-11-22
RD-484 Add labels to rest client

YoniItzhak:
Adding the `/labels` endpoint mechanism to the rest-client.

### DIFF
--- a/cloudify_rest_client/deployments.py
+++ b/cloudify_rest_client/deployments.py
@@ -361,7 +361,7 @@ class DeploymentsClient(object):
 
     def update_labels(self, deployment_id, labels):
         """
-        Updateas the deployment's labels.
+        Updates the deployment's labels.
 
         :param deployment_id: Deployment's id to update.
         :param labels: The new deployment's labels. A list of 1-entry
--- a/cloudify_rest_client/deployments.py
+++ b/cloudify_rest_client/deployments.py
@@ -16,6 +16,8 @@
 from cloudify_rest_client.responses import ListResponse
 from cloudify_rest_client.constants import VisibilityState
 
+from .labels import Label
+
 
 class Deployment(dict):
     """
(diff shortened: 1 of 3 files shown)


cloudify_rest_client/deployments.py
```diff
@@ -342,3 +358,17 @@ def set_site(self, deployment_id, site_name=None, detach_site=False):
             '/deployments/{0}/set-site'.format(deployment_id),
             data=data
         )
+
+    def update_labels(self, deployment_id, labels):
+        """
+        Updateas the deployment's labels.
```
tehasdf, line 364:
```suggestion
        Updates the deployment's labels.
```

cloudify_rest_client/labels.py
```diff
@@ -0,0 +1,45 @@
...
+        return self['creator_id']
+
+
+class DeploymentsLabelsClient(object):
+    def __init__(self, api):
+        self.api = api
+
+    def list_keys(self):
+        """
+        Retunrns a list of the deployments labels' keys in the current tenant.
```
tehasdf, line 32:
```suggestion
        Returns a list of the deployments labels' keys in the current tenant.
```

tehasdf: And tbh, you know that I like to drop as many words as I can get away with!

```suggestion
    All defined label keys, from all deployments.
```
Because
 - "Returns a list" is unnecessary, of course a method named "list keys" is going to return a list
 - "list" is doubly unnecessary, because what if it's not a list but a set? A set works even better. And btw, it's not even a list, it's a ListResponse
 - "in the current" tenant: all the restclient methods work in terms of the current tenant (modulo exceptions, with an explicit "all tenants", but still: current tenant is implicit even then), so no need to say that just here specifically

So we would maybe just drop the whole docstring then? But I feel it's important to underline that yes, that does look at all deployments

cloudify_rest_client/labels.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        response = self.api.get('/labels/deployments')
+        return ListResponse(response['items'], response['metadata'])
+
+    def list_key_values(self, label_key):
+        """
+        Returns a list of the deployments labels' values for the specified
```
tehasdf, line 39: also consider this, after considering the keys one

tehasdf (approved):

PR #600, opened 2020-11-29
RD-622 Add blueprint update endpoint

glukhman:


cloudify_rest_client/blueprints.py
```diff
@@ -404,3 +404,20 @@ def set_visibility(self, blueprint_id, visibility):
                 self=self, id=blueprint_id),
             data=data
         )
+
+    def update(self, blueprint_id, update_dict):
+        """
+        Update a blueprint.
+
+        Used for updating the blueprint's state (and error) while uploading,
+        And updating the blueprint's other attributes upon a successful upload.
```
YoniItzhak, line 413: And -> and

PR #602, opened 2020-12-08
Bump pyyaml to 5.3.1

tehasdf:
Py 2.6 doesn't support pyyaml 5.3.1. Therefore, specialcase it to keep 4.2.

The only place we still allow py 2.6 is the agent, and that does NOT
install the `.[dispatcher]` extra anyway. But we do want to be able to
still run unittests.

### DIFF
--- a/jenkins/Jenkinsfile
+++ b/jenkins/Jenkinsfile
@@ -141,8 +141,13 @@ pipeline {
                   fi'
                 pyenv install 2.6.9
                 pyenv global 2.6.9
-                '''    
-                install_test_dependencies('py26')
+                '''
+                // py26 uses py26-only-compatible versions of libs: pyyaml 4.2b4, hence the specialcased [dispatcher_py26]
+                sh '''#!/bin/bash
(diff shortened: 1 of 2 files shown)


tehasdf:
FWIW, this is because otherwise we can't install common in cloudify-manager and cloudify-premium (and specifically, their CI).
For reference, the error we get there is:
```
[2020-12-08T16:45:06.042Z] ERROR: Cannot install -r test-requirements.txt (line 2) and cloudify-common[dispatcher]==5.1.1.dev1 because these package versions have conflicting dependencies.
[2020-12-08T16:45:06.042Z] 
[2020-12-08T16:45:06.042Z] The conflict is caused by:
[2020-12-08T16:45:06.042Z]     cloudify-common[dispatcher] 5.1.1.dev1 depends on PyYAML==4.2b4
[2020-12-08T16:45:06.042Z]     cloudify-rest-service 5.1.1.dev1 depends on pyyaml==5.3.1
```

YoniItzhak (approved):

PR #612, opened 2021-01-01
RD-897 Give stdout+err on upgrade script failure

geokala:


geokala:
Note: This won't actually give output on failed agent upgrades until trying to upgrade to the next version, because upgrades are performed using the script_runner of the old agent.

tehasdf (approved):

tehasdf (approved):

PR #621, opened 2021-01-12
RD-695 Add more commands to filters

YoniItzhak:
Adding `get`, `delete` and `update` to filters

### DIFF
--- a/cloudify_rest_client/filters.py
+++ b/cloudify_rest_client/filters.py
@@ -1,5 +1,3 @@
-import json
-
 from cloudify_rest_client.responses import ListResponse
 from cloudify_rest_client.constants import VisibilityState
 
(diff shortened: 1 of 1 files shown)


cloudify_rest_client/filters.py
```diff
@@ -71,3 +75,34 @@ def list(self, **kwargs):
...
+        :param new_filter_rules: A new list of filter rules. Filter rules must
+               be one of: <key>=<value>, <key>=[<value1>,<value2>,...],
+               <key>!=<value>, <key>!=[<value1>,<value2>,...], <key> is null,
+               <key> is not null
+        :param new_visibility: The new visibility to update
+        :return: The updated filter
+        """
+        data = {}
+        if not new_filter_rules and not new_visibility:
+            raise RuntimeError('In order to update a filter, you must specify '
```
YoniItzhak, line 99: I think this should be the behavior, if no update is done, then throw an error. What do you think?

tehasdf: sure, i dont think this is very important

tehasdf (approved):

PR #632, opened 2021-01-18
RD-1054 Add exec schedules REST client

glukhman:


cloudify_rest_client/execution_schedules.py
```diff
@@ -0,0 +1,258 @@
+from cloudify_rest_client.responses import ListResponse
+
+
+class ExecutionSchedule(dict):
+    """
+    Cloudify snapshot.
```
tehasdf, line 6: nope :P

cloudify_rest_client/execution_schedules.py
```diff
@@ -0,0 +1,258 @@
...
+        return self.get('stop_on_fail')
+
+
+class ExecutionSchedulesClient(object):
+
+    def __init__(self, api):
+        self.api = api
+        self._uri_prefix = 'execution-schedules'
+
+    def create(self, schedule_id, deployment_id, workflow_id, parameters=None,
```
tehasdf, line 88: i would suggest grouping the parameters, and having the create method accept a "execution_arguments" or something dict, so that there's only 50 arguments to this function and not ten thousand

tehasdf (approved):

PR #673, opened 2021-02-22
RD-1185 Add deployment schedules of DSL parser

glukhman:


cloudify/utils.py
```diff
@@ -914,3 +917,74 @@ def target_plugin_prefix(name, tenant_name, version=None, deployment_id=None):
     # declared always take precedence
     parts += [name, version or '0.0.0']
     return os.path.join(*parts)
+
+
+def parse_utc_datetime_from_expression(time_expression, timezone):
```
tehasdf, line 922: what does it mean "from expression"? as opposed to what?

cloudify/utils.py
```diff
@@ -914,3 +917,74 @@ def target_plugin_prefix(name, tenant_name, version=None, deployment_id=None):
...
+                "{} is not a recognized timezone".format(timezone))
+        return pytz.timezone(timezone).localize(date_time).astimezone(
+            pytz.utc).replace(tzinfo=None)
+    else:
+        ts = time.time()
+        return date_time - (datetime.fromtimestamp(ts) -
+                            datetime.utcfromtimestamp(ts))
+
+
+def parse_timedelta(expr):
```
tehasdf, line 952: why "parse_timedelta" if this doesn't return a timedelta?

cloudify/utils.py
```diff
@@ -914,3 +917,74 @@ def target_plugin_prefix(name, tenant_name, version=None, deployment_id=None):
...
+    match = r"^(\d+)\ ?(min(ute)?|h(our)?|d(ay)?|w(eek)?|mo(nth)?|y(ear)?)s?$"
+    parsed = re.findall(match, expr)
+    if not parsed or len(parsed[0]) < 2:
+        raise NonRecoverableError("{} is not a legal time delta".format(expr))
+    return int(parsed[0][0]), parsed[0][1]
+
+
+def parse_and_apply_timedelta(expr, date_time):
```
tehasdf, line 960: "and" is very bad for functions, but thankfully there's already a separate "parse_timedelta" function that could do the parsing
then all the year/month/week/day/hour could go live in there

dsl_parser/elements/deployment_schedules.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def parse(self):
+        timezone = self.initial_value.get('timezone')
+        since = self.initial_value['since']
+        until = self.initial_value.get('until')
+        return_dict = self.initial_value.copy()
+
+        # calculate naive UTC datetime from time expressions in since and until
+        time_fmt = '%Y-%m-%d %H:%M:%S'
+        return_dict['since'] = datetime.strftime(
```
tehasdf, line 143: do we really need strings here?

dsl_parser/elements/deployment_schedules.py
```diff
@@ -0,0 +1,155 @@
...
+    schema = Leaf(type=dict)
+
+
+class Weekday(Element):
+    schema = Leaf(type=text_type)
+
+    def validate(self):
+        weekdays_list = ['su', 'mo', 'tu', 'we', 'th', 'fr', 'sa']
+        err_msg = "{0} is not a valid weekday value. Accepted values: {1}"
+        if self.initial_value not in weekdays_list:
```
tehasdf, line 102: lower, perhaps

dsl_parser/elements/deployment_schedules.py
```diff
@@ -0,0 +1,155 @@
...
+                                           Dict)
+from cloudify.utils import parse_utc_datetime_from_expression, parse_timedelta
+
+
+class ScheduleEnabled(Element):
+    schema = Leaf(type=bool)
+
+    def parse(self):
+        if not self.initial_value:
+            return True
```
tehasdf, line 19: and otherwise return... nothing?

tehasdf (changes requested):
1. the parse methods need _at least_ docstrings
2. they also need unittests
3. the dsl needs tests as well

cloudify/utils.py
glukhman: well, the problem here is that datetime.timedelta doesn't actually handle months and years, so this requires that kind of hybrid function

dsl_parser/elements/deployment_schedules.py
glukhman: yes, datetime objects aren't JSON-serializable

tehasdf (approved):

PR #685, opened 2021-02-25
Rd 1264 rollback workflow

AdarShaked:


tehasdf (commented):
i guess the workflow makes sense, but really needs the docstring to explain it better, because what is a "resolved" state? from reading the docstring, the user needs to be able to understand that "from state X, we move to state Y"

and yes tests would be useful, when you write the "we move from state X to Y" declarations - that's also the list of cases to test

PR #693, opened 2021-03-02
RD-1554/1555 add LDAP configuration service task

geokala:


geokala:
Not required, approach changed.

PR #705, opened 2021-03-09
Revert "Temporarily modify pyenv source repo"

mateumann:
Reverts cloudify-cosmo/cloudify-common#704


geokala (approved):

geokala:
For any later scholars of history:This commit and the one it reverted were due to a pyenv bug that they merged the fix for faster than we expected.

PR #709, opened 2021-03-10
RD-1655 depgroup create: allow providing more than just inputs

tehasdf:
When creating new deployments via group, instead of providing just inputs,
allow providing a whole deployment dict, and new deployments will be based
off of that.

These dicts, which specify new deployments to be created, can contain
so far just the key "inputs", but later in RD-1655 we'll also allow providing
things like the id, deployment_settings, labels, etc

### DIFF
--- a/cloudify_rest_client/deployments.py
+++ b/cloudify_rest_client/deployments.py
@@ -256,7 +256,8 @@ class DeploymentGroupsClient(object):
         :param description: description of the group
         :param blueprint_id: the default blueprint to use when extending
         :param default_inputs: the default inputs to use when extending
-        :param deployment_ids: set the group deployments to these
+        :param deployment_ids: make the group contain these
+                               existing deployments
         :param filter_id: set the group to contain the deployments matching
                           this filter
         :param inputs: create new deployments using these inputs merged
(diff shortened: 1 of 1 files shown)


EarthmanT (approved):

cloudify_rest_client/deployments.py
```diff
@@ -256,13 +256,21 @@ def put(self, group_id, visibility=VisibilityState.TENANT,
...
         :param filter_id: set the group to contain the deployments matching
-                          this filter
-        :param inputs: create new deployments using these inputs merged
-                       with default_inputs, and using default_blueprint
+            this filter
+        :param new_deployments: create new deployments using this
+            specification, merged with the group's default_blueprint
+            and default_inputs
+        :type new_deployments: a list of dicts, each can contain the
+            key "inputs"
```
geokala, line 267: ... which will be used to populate the inputs of the new deployment?

geokala (approved):

cloudify_rest_client/deployments.py
geokala: After some discussion, I think we should stick with this for the moment and see if we need to get more descriptive when more fields added.

PR #729, opened 2021-03-22
RD-1485 Update deployment client response

mabuaisha:
This PR created in order to do the following:
1. Include extra params response as part of deployment in client deployments
2. Add new exception type when deployment reference something that does not exist

### DIFF
--- a/cloudify_rest_client/exceptions.py
+++ b/cloudify_rest_client/exceptions.py
@@ -281,6 +281,14 @@ class InvalidFilterRule(CloudifyClientError):
         self.err_reason = response.json().get('err_reason')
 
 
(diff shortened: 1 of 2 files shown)


cloudify_rest_client/exceptions.py
```diff
@@ -313,4 +321,5 @@ def __init__(self, message, server_traceback=None,
         MissingCloudifyLicense,
         ExpiredCloudifyLicense,
         InvalidFilterRule,
+        DeploymentParentNotFound
```
tehasdf, line 324: plz keep the trailing comma

mabuaisha: Fixed

geokala (approved):

PR #765, opened 2021-05-10
RD-2288 Remove useless retries

tehasdf:
I don't know why was this request retried. No other request is
retried like this.

All this does, is makes a 401 get displayed multiple times,
and in case a 500 happens, it will be hidden by the additional 401s.

This dates back to CFY-1599, which was related to cancelling, but
I dont see how this specifically relates. Anyway, CFY-1599 didn't
really solve the cancel-in-pending issue, only by the time we
introduced kill-cancels that was really fixed.


tehasdf:
ah, last commit's message is kinda wrong, it did actually have `InvalidExecutionUpdateStatus` as the last retry

geokala (approved):

PR #771, opened 2021-05-12
RD-2015 Fix condition in for namespaced workflows

glukhman:
if `_is_remote_script_resource(operation_mapping, remote_resources_namespaces)` returns true, but we actually have a plugin to run the operation, we don't want to default to script plugin running it, but instead it should continue to the other branch, under `if candidate_plugins`

### DIFF
--- a/dsl_parser/elements/operation.py
+++ b/dsl_parser/elements/operation.py
@@ -297,8 +297,9 @@ def process_operation(
 
     if (utils.is_valid_url(operation_mapping) or
         _is_local_script_resource_exists(resource_bases, operation_mapping) or
-        _is_remote_script_resource(operation_mapping,
-                                   remote_resources_namespaces)):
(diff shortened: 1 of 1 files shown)


dsl_parser/elements/operation.py
```diff
@@ -297,8 +297,9 @@ def process_operation(
 
     if (utils.is_valid_url(operation_mapping) or
         _is_local_script_resource_exists(resource_bases, operation_mapping) or
-        _is_remote_script_resource(operation_mapping,
-                                   remote_resources_namespaces)):
+        (not candidate_plugins and _is_remote_script_resource(
```
tehasdf, line 300: why?

tehasdf (changes requested):

tehasdf (approved):

PR #782, opened 2021-05-19
Avoid printing out the plugin dict, in plugin installation errors

tehasdf:
The plugin representation contains the error, so don't put it in
errors, otherwise we get the error in the error, and that quickly
leads to the string getting jsonified (and escaped) fifty times


geokala:
Or even 255 times. It eventually stops when the mgmtworker chokes!

geokala (approved):

PR #794, opened 2021-06-03
RD-2509 Speed up parsing with faster sibling check

geokala:


geokala:
Parse times before these changes:
awsvm.yaml: 19.29  # The system tests aws/vm.yaml
just_dsl_version.yaml: 0.00
just_types.yaml: 0.40
openstackvm.yaml: 2.67  # The system tests openstack/vm.yaml

All plugin.yaml and types.yaml files were stored locally to avoid network latency.

With these changes:
awsvm.yaml: 3.06
just_dsl_version.yaml: 0.00
just_types.yaml: 0.25
openstackvm.yaml: 1.23

dsl_parser/framework/parser.py
```diff
@@ -439,6 +439,22 @@ def _calculate_element_graph(self):
                         self.element_graph.add_edge(element, dep)
                     continue
 
+                if predicates == [sibling_predicate]:
+                    # These appear to be generally in order, which means
+                    # that we can cut parsing time by a massive amount for
+                    # plugins which have a lot of DSL elements.
```
tehasdf, line 445: no reason to mention plugins here, this parses blueprints in general and a "plugin" isn't really a thing at this level of abstraction

dsl_parser/framework/parser.py
```diff
@@ -439,6 +439,22 @@ def _calculate_element_graph(self):
                         self.element_graph.add_edge(element, dep)
                     continue
 
+                if predicates == [sibling_predicate]:
+                    # These appear to be generally in order, which means
```
tehasdf, line 443: `These`? predicates? elements?
and what being in order (what order, anyway?) has to do with anything?

dsl_parser/framework/parser.py
```diff
@@ -439,6 +439,22 @@ def _calculate_element_graph(self):
...
+                    # If we don't do this, our time complexity is n**2 as
+                    # we compare all 'default' elements to all 'type'
+                    # elements (for example), when all we care about is if
+                    # they are siblings.
+                    for dependency in dependencies:
+                        for element in dependency.parent().children():
+                            if element in _elements:
+                                self.element_graph.add_edge(element,
+                                                            dependency)
+                                assert predicates[0](element, dependency)
```
tehasdf, line 455: this seems useless, surely we don't need to assert that networkx works correctly

dsl_parser/framework/parser.py
```diff
@@ -439,6 +439,22 @@ def _calculate_element_graph(self):
...
+                    continue
+
                 for dependency in dependencies:
                     for element in _elements:
                         add_dependency = all([
```
tehasdf, line 460: while you're at it, feel free to drop the pointless `[]` inside of the `all()` call, which also saves some... small amount of time

dsl_parser/framework/parser.py
```diff
@@ -596,8 +612,25 @@ def _extract_element_requirements(element):
...
+                    # about 7 seconds on sufficiently large blueprints (e.g.
+                    # ones that use the AWS plugin).
+                    required_type_elements = [
+                        child for child in element.parent().children()
+                        if isinstance(child, required_type)]
+                else:
+                    required_type_elements = (
+                        context.element_type_to_elements.get(required_type,
+                                                             [])
+                    )
```
tehasdf, line 632: hah now that's some weird style, to avoid a `\`

dsl_parser/framework/parser.py
```diff
@@ -439,6 +439,22 @@ def _calculate_element_graph(self):
...
+                    # These appear to be generally in order, which means
+                    # that we can cut parsing time by a massive amount for
+                    # plugins which have a lot of DSL elements.
+                    # If we don't do this, our time complexity is n**2 as
+                    # we compare all 'default' elements to all 'type'
+                    # elements (for example), when all we care about is if
+                    # they are siblings.
+                    for dependency in dependencies:
+                        for element in dependency.parent().children():
+                            if element in _elements:
```
tehasdf, line 452: `set() & set()` would work well here, although it doesn't really speed it up by much.
But it seems that in general `self.element_type_to_elements` would be better as a dict of sets, rather than a dict of lists as it currently is.

dsl_parser/framework/parser.py
geokala: Oh right, that comment was from a previous shot at the logic, I'll update.

dsl_parser/framework/parser.py
geokala: Oh, sorry- that was when I was testing it.

dsl_parser/framework/parser.py
geokala: Comments updated.

dsl_parser/framework/parser.py
geokala: Comments updated.

dsl_parser/framework/parser.py
geokala: assert removed.

dsl_parser/framework/parser.py
geokala: Updated.

dsl_parser/framework/parser.py
geokala: It's very pretty!

tehasdf (approved):

dsl_parser/framework/parser.py
geokala: Discussed via slack- verdict: Not worth the extra time/investigation today.

PR #797, opened 2021-06-08
RD-2579 _SetNodeInstanceStateTask: force-set the state

tehasdf:
We used to first fetch the instance, to figure out the current
version, and then set the new state anyway.
So, the only reason we fetched it, was so that our update always
succeeds - we didn't react to the received state in any way.
We can simply just cut out the middleman, and force-set the
state directly, without fetching it.

And anyway, force-setting, like we (essentially) did before,
and do now, does make sense for the state - instance state is never
changed concurrently, only runtime-properties are (and that's what
version is for).

### DIFF
--- a/cloudify/workflows/tasks.py
+++ b/cloudify/workflows/tasks.py
@@ -25,7 +25,6 @@ from cloudify.manager import (
     get_rest_client,
     get_node_instance,
     update_execution_status,
-    update_node_instance
 )
 from cloudify.constants import (
     MGMTWORKER_QUEUE,
@@ -976,10 +975,11 @@ class _SetNodeInstanceStateTask(_LocalTask):
         }
 
     def remote(self):
-        node_instance = get_node_instance(self._node_instance_id)
-        node_instance.state = self._state
-        update_node_instance(node_instance)
-        return node_instance
+        get_rest_client().node_instances.update(
+            self._node_instance_id,
+            force=True,
+            state=self._state
+        )
 
     def local(self):
         self.storage.update_node_instance(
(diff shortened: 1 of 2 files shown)


cloudify/workflows/tasks.py
```diff
@@ -976,10 +975,11 @@ def dump(self):
...
     def remote(self):
-        node_instance = get_node_instance(self._node_instance_id)
-        node_instance.state = self._state
-        update_node_instance(node_instance)
-        return node_instance
+        get_rest_client().node_instances.update(
+            self._node_instance_id,
+            force=True,
+            state=self._state
+        )
```
geokala, line 982: Won't this result in an instance whose state is updated having its version reset to 1?

tehasdf: nope! we never set the version field ourselves - it's actually a sqlalchemy built-in feature, and the version field is incremented with any update, https://github.com/cloudify-cosmo/cloudify-manager/blob/bd26be23cd7313584aa5175da4c4a1775410677f/rest-service/manager_rest/storage/resource_models.py#L1765-L1766

tehasdf: you'll also notice that in the endpoint, the version from the request is used _only_ for the comparison: https://github.com/cloudify-cosmo/cloudify-manager/blob/bd26be23cd7313584aa5175da4c4a1775410677f/rest-service/manager_rest/rest/resources_v1/nodes.py#L210-L228

geokala: OK- makes our interface feel a bit misleading but that's not something I think we should worry about as part of this.

geokala (approved):

PR #805, opened 2021-06-18
RD-515 Don't create events for operations

tehasdf:
They will be automatically created when updating the operation

### DIFF
--- a/cloudify/workflows/events.py
+++ b/cloudify/workflows/events.py
@@ -14,7 +14,13 @@
 #    * limitations under the License.
 
 from cloudify import logs, utils
-from cloudify.workflows import tasks as tasks_api
+from cloudify.constants import (
+    TASK_SENDING,
+    TASK_STARTED,
+    TASK_RESCHEDULED,
+    TASK_SUCCEEDED,
+    TASK_FAILED,
+)
 
 
 def send_task_event_func_remote(task, event_type, message,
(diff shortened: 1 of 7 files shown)


cloudify/workflows/events.py
```diff
@@ -48,7 +54,54 @@ def _send_task_event_func(task, event_type, message, out_func,
 
 
 def _filter_task(task, state):
-    return state != tasks_api.TASK_FAILED and not task.send_task_events
+    return state != TASK_FAILED and not task.send_task_events
+
+
+def get_event_type(state):
+    if state == TASK_SENDING:
```
geokala, line 61: A dict might be better here- it'll be shorter as more things are added, and could be reused in other places if it needed to be.

tehasdf: yeah, I thought about that. But then I decided I preferred this, so that it stays similar to the other function. wdyt?

geokala: Assuming you mean format_event_message, I was looking at that and wondering if we can maybe do something better since both of these seem to have (aside from sending) very similar structures.

geokala: e.g. here we could do something like:
if state == TASK_SENDING: return 'sending_task'
elif state in (TASK_STARTED, TASK_SUCCEEDED, TASK_RESCHEDULED, TASK_FAILED): return 'task_{}'.format(state)
else: raise...

The format_event_message could do something similar for the basic message formatting- message = "Task {} '{{0}}'".format(state)
That'll be a few reduced lines and make the special cases look more special?

geokala (commented):
Is it likely to be a back compat issue that we've dropped one local task event? That's the 'task sent' event, right? Could users reasonably be watching for that in order to do something?

tehasdf:
@geokala perhaps. But that's for local events. I'm not worried about keeping compat for `cfy local`.

And for the other-kind of local events, well, surely you're not watching for events of `_SetNodeInstanceStateTask` because there are none anyway :)

geokala:
Fair- then I'm not worried about the back compat aspect there, given the current state of cfy local.

geokala (approved):

PR #806, opened 2021-06-22
RD-518 Insert logs/events via a REST call

tehasdf:
Instead of sending in events via AMQP, create them via REST.

Yes, this means amqp-postgres can be removed!


tehasdf:
this is based off of the other branch, but after that is merged, I will change this to be based off of master

geokala (approved):

PR #812, opened 2021-06-25
RD-2650 Don't explicitly insert workflow node events

tehasdf:
Those will be inserted automatically, saving a request


tehasdf:
the part that stores the event is cloudify-cosmo/cloudify-manager#3128

geokala (approved):

PR #832, opened 2021-07-13
update_operation: pass through exception_causes

tehasdf:
well if we str() the exception and then getattr causes, then obviously
the causes is None, because we're getattr-ing off of the string!
instead, let's getattr from the actual exception.

I could've just as well simply reversed the order of the two lines,
but this seems better


tehasdf:
yes i performed an actual proper facepalm irl after finding this one

tehasdf:
this fixes `integration_tests/tests/agentless_tests/test_execution_logging.py::ExecutionLoggingTest::test_execution_logging`

mateumann (approved):

PR #843, opened 2021-07-30
BlockinReqRespHandler: synchronously delete the queue

tehasdf:
We'd do good to delete this response queue synchronously, otherwise
it might not actually get deleted if we close the process before
the delete request gets sent


geokala (approved):

tehasdf:
no, this can't be done this way

PR #846, opened 2021-08-11
RD-2587 optimize pipeline

yuvlar:
Made the pod burstable cpu-wise so the pytest will utilize the most out of the resources.


jenkins/Jenkinsfile
```diff
@@ -63,29 +63,15 @@ pipeline {
             sh script: "mkdir -p ${env.WORKSPACE}/py3_compat && cp -rf ${env.WORKSPACE}/${env.PROJECT}/. ${env.WORKSPACE}/py3_compat", label: "copying repo to seperate workspace"
             container('py27'){
               dir("${env.WORKSPACE}/py3_compat"){
-                echo 'install futurize, remove compat modules and find python3-incompatible code'
+                echo 'Prepare for py3compat'
```
tehasdf, line 66: you can just use a `label` for the `sh`

jenkins/build-pod.yaml
```diff
@@ -2,6 +2,27 @@ apiVersion: v1
 kind: Pod
 spec:
   containers:
+  - name: jnlp
+    image: jenkins/inbound-agent:4.3-4
+    resources:
+      limits:
+        cpu: 0.2
+        memory: 256Mi
```
tehasdf, line 10: do we even need this?

tehasdf (approved):

jenkins/Jenkinsfile
yuvlar: True but I forget about it sometimes. I know it doesn't do any actual change in the performance so I'll skip it this time, but will add it next time

PR #856, opened 2021-08-19
RD-3008 Define operations to store in audit_log

mateumann:


mateumann:
definitions 100% in cloudify-manager

PR #866, opened 2021-08-31
RD-3004 get attributes list intrinsic function

geokala:


tehasdf (dismissed):

dsl_parser/tests/test_get_attributes_list.py
```diff
@@ -0,0 +1,263 @@
...
+            {'id': 'badnode'},
+            {'id': 'goodnode', 'properties': {'sing': 'give'}},
+        ]
+        storage = self._mock_evaluation_storage(
+            node_instances=node_instances, nodes=nodes)
+
+        payload = {'a': {'get_attributes_list': ['goodnode', 'sing']}}
+        context = {}
+        result = functions.evaluate_functions(payload.copy(), context, storage)
+        print(result)
```
mateumann, line 119: drop this line, please

dsl_parser/functions.py
```diff
@@ -592,6 +547,103 @@ def _validate_ref(self, ref, ref_name):
...
 
+def _get_attribute_from_node_instance(ni, node, attribute_path, path):
+    value = _get_property_value(ni['node_id'],
+                                ni.get('runtime_properties'),
+                                attribute_path,
+                                path,
+                                raise_if_not_found=False)
+    if value is None:
+        # attribute not found in instance runtime properties
+        if value is None:
```
tehasdf, line 558: eh?

dsl_parser/functions.py
```diff
@@ -592,6 +547,103 @@ def _validate_ref(self, ref, ref_name):
...
+                         '{2}.'.format(node_name,
+                                       name,
+                                       path))
+    if node_name not in [SELF, SOURCE, TARGET]:
+        found = [
+            x for x in plan.node_templates if node_name == x['id']]
+        if not found:
+            raise KeyError(
+                "{0} function node reference '{1}' does not exist.".format(
+                    name, node_name))
```
tehasdf, line 606:
```suggestion
        if not any(node_name == template['id']
                        for template in plan.node_templates):
            raise KeyError(
                "{0} function node reference '{1}' does not exist.".format(
                    name, node_name))
```
no need to store a list, if we're only checking if any item exists

dsl_parser/functions.py
geokala: Whoops, fixing

dsl_parser/tests/test_get_attributes_list.py
```diff
@@ -0,0 +1,263 @@
...
+            },
+        ]
+        nodes = [
+            {'id': 'badnode'},
+        ]
+        storage = self._mock_evaluation_storage(node_instances, nodes)
+        payload = {'a': {'get_attributes_list': ['goodnode', 'sing']}}
+        with testtools.testcase.ExpectedException(
+                KeyError,
+                '.*Node not found.*'):
```
mateumann, line 234: Why not just `Node not found`?

dsl_parser/functions.py
```diff
@@ -592,6 +547,103 @@ def _validate_ref(self, ref, ref_name):
...
+        raise ValueError('{0} cannot be used with {1} function in '
+                         '{2}.'.format(node_name,
+                                       name,
+                                       path))
+    if scope == scan.NODE_TEMPLATE_RELATIONSHIP_SCOPE and \
+            node_name == SELF:
+        raise ValueError('{0} cannot be used with {1} function in '
+                         '{2}.'.format(node_name,
+                                       name,
+                                       path))
```
tehasdf, line 599: perhaps we can DRY this thing, because the very same error is raised 3 times in this function. A more involved `if`, to put all these conditions into a single `or`, perhaps?

dsl_parser/functions.py
```diff
@@ -592,6 +547,103 @@ def _validate_ref(self, ref, ref_name):
...
+                                raise_if_not_found=False)
+    if value is None:
+        # attribute not found in instance runtime properties
+        if value is None:
+            # special case for { get_attributes_list: [
+            #                        ..., node_instance_id] }
+            # returns the node-instance-id
+            if len(attribute_path) == 1\
+                    and attribute_path[0] == 'node_instance_id':
+                value = ni['id']
```
tehasdf: should this be here, or before checking the runtime props?

dsl_parser/functions.py
geokala: That's where was refactoring, must've typed one out then copied the other in, fixed.

dsl_parser/functions.py
geokala: Fair point- that was in the original code but agreed it's unnecessary, I'll clear it.

dsl_parser/tests/test_get_attributes_list.py
```diff
@@ -0,0 +1,262 @@
...
+            node_instances=node_instances, nodes=nodes)
+
+        payload = {'a': {'get_attributes_list': ['goodnode', 'sing']}}
+        context = {}
+        result = functions.evaluate_functions(payload.copy(), context, storage)
+
+        attributes_list = result['a']
+        assert len(attributes_list) == 5
+        for value in ('never', 'gonna', 'give', 'you', 'up'):
+            assert value in attributes_list
```
tehasdf, line 123: if attributes_list is `['up', 'up', 'up', 'up', 'up']` then this test will pass, while it shouldn't have.
I suggest either set()==set(), or just a straight list comparison, which would also assert on the ordering, and I suppose we might as well guarantee ordering within this function?

dsl_parser/functions.py
geokala: Fixed.

dsl_parser/tests/test_get_attributes_list.py
```diff
@@ -0,0 +1,262 @@
...
+            },
+        ]
+        nodes = [
+            {'id': 'badnode'},
+        ]
+        storage = self._mock_evaluation_storage(node_instances, nodes)
+        payload = {'a': {'get_attributes_list': ['goodnode', 'sing']}}
+        with testtools.testcase.ExpectedException(
+                KeyError,
+                '.*Node not found.*'):
```
tehasdf, line 234: I'd expect the `.*` at the beginning and end are implied, at least that's how other raisesRegexp methods work?

dsl_parser/functions.py
geokala: It should be there- it appears we made it a pseudo-runtime property.

This means we go:
1. Is it in the runtime properties? If yes, use it.
2. Otherwse: Is it called node_instance_id? Then return the id of the node_instance.
3. Otherwise: If it's in node properties, return it
4. Otherwise, return nothing

dsl_parser/tests/test_get_attributes_list.py
tehasdf: Actually, I'm wrong, the loop is the other way around, you're right! Still, might want to assert on the ordering anyway?

geokala: True, I'll change it to an all check instead.

I don't think we can guarantee the ordering can we? Under the hood we're retrieving the node_instances list, and I don't think the order is deterministic.

geokala: Switched to an all() check.

dsl_parser/tests/test_get_attributes_list.py
geokala: It's not in this one, as I discovered earlier today.

dsl_parser/tests/test_get_attributes_list.py
tehasdf: But within this function, we can, ie. the guarantee is that "this function returns whatever ordering the backend gives us". I do expect it'll become a question with these functions, as this is pretty much the first list-ordering function we have. So just having the order depend on one thing (essentially, storage-manager) (who does sort by default, on... something. Storage id?) is not bad

geokala: That's a fair point- OK, I'll change it to that and add a comment to explain that.

dsl_parser/tests/test_get_attributes_list.py
geokala: Because it's actually preceded by a quote and this regex part doesn't implicitly prefix with .*

tehasdf (approved):

PR #870, opened 2021-09-01
rd-3005 get_attributes_dict

geokala:


dsl_parser/functions.py
```diff
@@ -633,6 +633,79 @@ def evaluate(self, handler):
...
+            for attribute_path in self.attribute_paths
+        ]
+        duplicated_or_ambiguous = [
+            item for item in duplicate_or_ambiguous_check
+            if duplicate_or_ambiguous_check.count(item) > 1
+        ]
+        if duplicated_or_ambiguous:
+            raise exceptions.FunctionEvaluationError(
+                'One or more duplicated or ambiguous attribute paths found: '
+                '{}'.format(', '.join(duplicated_or_ambiguous))
```
tehasdf, line 674: 2.6 for ci?

dsl_parser/functions.py
```diff
@@ -633,6 +633,79 @@ def evaluate(self, handler):
...
+                    'but got: {1}.'.format(self.name, args))
+
+        duplicate_or_ambiguous_check = [
+            self._convert_to_identifier(attribute_path)
+            for attribute_path in self.attribute_paths
+        ]
+        duplicated_or_ambiguous = [
+            item for item in duplicate_or_ambiguous_check
+            if duplicate_or_ambiguous_check.count(item) > 1
+        ]
```
tehasdf, line 670: i realize this is going to be short, but as a rule, i don't really want quadratic code in here. Maybe a dict counter?

dsl_parser/functions.py
```diff
@@ -633,6 +633,79 @@ def evaluate(self, handler):
...
+            _validate_ref(node_id, self.node_name, self.name,
+                          self.path, self.attribute_paths)
+        else:
+            node_id = self.node_name
+        node = handler.get_node(node_id)
+        node_instances = handler.get_node_instances(node_id)
+
+        results = {}
+
+        for ni in node_instances:
```
tehasdf, line 693: ni!

dsl_parser/functions.py
```diff
@@ -633,6 +633,79 @@ def evaluate(self, handler):
...
+class GetAttributesDict(Function):
+
+    def __init__(self, args, **kwargs):
+        self.node_name = None
+        self.attribute_path = None
+        super(GetAttributesDict, self).__init__(args, **kwargs)
+
+    def parse_args(self, args):
+        self.node_name = args[0]
+        self.attribute_paths = args[1:]
```
tehasdf, line 646: this intrinsic function syntax is very ugly, because it's `{get_attributes_dict: [node, attr1, attr2]}`, so we have a list that is nowhere near homogenous, because it has both the node name, and the attr names. Wouldn't it be better to do `{get_attributes_dict: [node, [attr1, attr2]]}` at least?

dsl_parser/functions.py
geokala: Actually just passed, strangely enough. Given that we want to remove 2.6 for this anyway, is it worth changing?

dsl_parser/functions.py
geokala: Does that actually avoid being quadratic under the hood?

dsl_parser/functions.py
geokala: Your shrubbery is out for delivery with our driver: Arthur.

dsl_parser/functions.py
tehasdf: huh. Then no.

dsl_parser/functions.py
tehasdf: yes, the issue is count() (which is O(N)) inside of the O(N) loop. Putting them in a dict would be O(N) time + O(N) memory. Of course with a constant for the time, because you'd need one loop to put them in the dict, and then another loop to go over the dict to check the counts.

geokala: Ah, I see what you mean- I originally started coding it with an extra counter-dict step, but for some reason thought this looked more elegant despite the unease of the quadratic. I'll tweak it.

tehasdf (approved):

dsl_parser/functions.py
geokala: After a bit of further discussion on slack, we have decided to leave this using the current syntax as it's closer to the other get_attribute functions

PR #877, opened 2021-09-09
RD-3125 Retrieve relationship from node

mateumann:
Node instances which have two or more relationships to the same target
failed to update because only one of these relationships is stored for
the `node_instance` in DB.  This patch provides a fail-over mechanism
for such cases and build relationships information based on a relevant
node's relationships.

### DIFF
--- a/dsl_parser/rel_graph.py
+++ b/dsl_parser/rel_graph.py
@@ -595,10 +595,10 @@ def _extract_contained(node, node_instance):
     # node_instance['relationships'] contains only one of them, which makes
     # this function always fail for those node_instances.  Let's try to create
     # a relationship dictionary based on node['relationships'] in this case.
-    for node_relationship in node['relationships']:
-        if (node_relationship['type'] ==
-                contained_node_relationship['type']):
-            if rel_node_instance_id:
+    if rel_node_instance_id:
+        for node_relationship in node['relationships']:
+            if (node_relationship['type'] ==
+                    contained_node_relationship['type']):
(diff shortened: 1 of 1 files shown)


dsl_parser/rel_graph.py
```diff
@@ -585,10 +585,23 @@ def _extract_contained(node, node_instance):
...
         if (node_instance_relationship['type'] ==
                 contained_node_relationship['type']):
             return node_instance_relationship
+    # In case there are two or more relationships to the same target,
+    # node_instance['relationships'] contains only one of them, which makes
+    # this function always fail for those node_instances.  Let's try to create
+    # a relationship dictionary based on node['relationships'] in this case.
+    for node_relationship in node['relationships']:
+        if (node_relationship['type'] ==
+                contained_node_relationship['type']):
```
geokala, line 600: Don't you need to assign rel_node_instance_id after this, or at least pass the thing that would've been assigned to the function?

geokala: Oh, no, I see what's happening. Ignore that.

dsl_parser/rel_graph.py
```diff
@@ -585,10 +585,23 @@ def _extract_contained(node, node_instance):
...
+            if rel_node_instance_id:
```
geokala, line 601: Why do you need to check for rel_node_instance_id in here? Surely if there were need for a check for it, it should be before you iterate over node relationships?

geokala (approved):

PR #887, opened 2021-09-20
RD-3228 restclient: expose the async_create flag in client.deps.create

tehasdf:
It's a feature that we have, and that I want to use, give me the
param.

### DIFF
--- a/cloudify_rest_client/deployments.py
+++ b/cloudify_rest_client/deployments.py
@@ -636,6 +636,7 @@ class DeploymentsClient(object):
         uri = '/deployments/{0}'.format(deployment_id)
         params = {}
         if async_create is not None:
+            # if it's None, we just keep the server's default behaviour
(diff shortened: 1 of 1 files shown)


mateumann (dismissed):

cloudify_rest_client/deployments.py
```diff
@@ -594,7 +594,8 @@ def create(self,
                site_name=None,
                runtime_only_evaluation=False,
                labels=None,
-               display_name=None):
+               display_name=None,
+               async_create=None):
```
geokala, line 598: Can't you default it to False and then just pass it in irrespective of whether it's not None later?

tehasdf: I don't want to do this, because that would then mean the new client is unable to work with pre-6.0 servers, and there's no need to do that

geokala (approved):

PR #892, opened 2021-09-27
RD-3252 Clean up query parameters for aiohttp get

mateumann:


cloudify_async_client/client.py
```diff
@@ -27,9 +27,19 @@ def url(self):
...
         timeout = kwargs.pop('timeout', 300)
+        if timeout is None:
+            timeout = 300
         if isinstance(timeout, int) or isinstance(timeout, float):
             timeout = aiohttp.ClientTimeout(total=timeout)
+
+        # Format query parameters
+        params = kwargs.copy()
+        p = {k: str(v) for k, v in params.pop('params', {}).items()
+             if v is not None}
```
tehasdf, line 38: this kinda sucks, we use `params` too many times. I suggest `def get(self, url, params=None, **kwargs):`, and then `if params: params={k:str(v) for k, v}`, and then `self.session.get(..., params=params, **kwargs)`

tehasdf: which, actually _do_ we need the str()?

mateumann: we do, because somewhere in the depths of aiohttp's ClientSession there is type-checking on the parameters' values

tehasdf (approved):

cloudify_async_client/client.py
```diff
@@ -25,10 +25,19 @@ def url(self):
         return '{0}://{1}:{2}/api/{3}'.format(self.protocol, self.host,
                                               self.port, self.api_version)
 
-    def get(self, url, **kwargs):
+    def get(self, url, params=None, **kwargs):
         timeout = kwargs.pop('timeout', 300)
```
geokala, line 29: Might as well just make this a defaulted arg?

cloudify_async_client/client.py
```diff
@@ -25,10 +25,19 @@ def url(self):
...
+        if timeout is None:
+            timeout = 300
         if isinstance(timeout, int) or isinstance(timeout, float):
             timeout = aiohttp.ClientTimeout(total=timeout)
```
geokala, line 33: Also, maybe just call aiohttp.ClientTimeout on it here- I'm assuming that we don't really support timeout being anything that isn't an integer or a float (do we?)

mateumann: I'd like to make it still possible to pass `aiohttp.ClientTimeout` as a `timeout` parameter (in case someone wants to fine-tune the specific timeouts).

```suggestion
 def get(self, url, params=None, timeout=300, **kwargs):
        if isinstance(timeout, int) or isinstance(timeout, float):
            timeout = aiohttp.ClientTimeout(total=timeout)
```

geokala: Ah, I see your reasoning- yeah, makes sense.

geokala (approved):

PR #895, opened 2021-10-03
test_operation_retry: don't wait 3 seconds in the test

tehasdf:
no reason to just sleep for 3 seconds. The tests themselves do pass
`task_retry_interval=0`, but #845 made it so that the retry_after
passed into ctx.operation.retry() is actually not ignored. So it
made tests just sleep for no reason. So let's not.


tehasdf:
this makes the tests some 20-odd seconds quicker to run

mateumann (approved):

PR #929, opened 2021-11-30
RD-3625 `get_results` intrinsic function

mateumann:
This is a "skeleton" which is goint to be filled-in soon with some
actual functions' implementations.


tehasdf (changes requested):
Instead, how about separate `string_split`, `string_capitalize`, etc, functions

mateumann:
We've decided to go with separate `string_…` intrinsic functions

PR #939, opened 2021-12-08
RD-3716 string_find DSL instrinsic function

mateumann:


dsl_parser/functions.py
```diff
@@ -1072,6 +1072,46 @@ def evaluate(self, handler):
...
 
+@register(name='string_find', func_eval_type=HYBRID_FUNC)
+class StringFind(Function):
+    def __init__(self, args, **kwargs):
+        self.haystack = None
+        self.needle = None
+        super(StringFind, self).__init__(args, **kwargs)
+
+    def parse_args(self, args):
+        if len(args) < 2:
```
tehasdf, line 1083:
```suggestion
        if len(args) != 2:
```
error message says "exactly two" so let's make it exactly two :P

tehasdf (dismissed):

tehasdf (approved):

PR #955, opened 2021-12-17
RD-3675 Add the check_status workflow

tehasdf:
The workflow is simple enough, but do see commits separately.
Anyway, it's very similar to execute_operation, but it will only run
the one specific task, and it will also store the results in
system_properties.

For now, the results are just the boolean "ok". We'll be adding more
data in there soon.

### DIFF
--- a/cloudify/tests/test_builtin_workflows.py
+++ b/cloudify/tests/test_builtin_workflows.py
@@ -15,6 +15,7 @@
 
 
 import time
+import unittest
 from os import path
 
 import mock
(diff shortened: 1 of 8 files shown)


cloudify/tests/test_builtin_workflows.py
```diff
@@ -660,6 +661,68 @@ def _get_node_instance(self, node_id):
...
+        assert len(noop_instances) == 1
+
+        pass_instance = pass_instances[0]
+        assert pass_instance['system_properties']
+        assert pass_instance['system_properties']['ok']
+
+        # the noop instance did run, even though it depends on a failing one
+        noop_instance = noop_instances[0]
+        assert noop_instance['system_properties']
+        assert noop_instance['system_properties']['ok']
```
mateumann, line 691: why should it be ok?

tehasdf: I declare that instances who have no check defined at all, are considered green

mateumann (approved):

PR #967, opened 2022-01-05
RD-3748 Pass tenant_name when creating IDDs

mateumann:
... because of times when multiple tenants use the same deployment_id
for different deployment and we should be able to find out which one
should be used.


cloudify_rest_client/inter_deployment_dependencies.py
```diff
@@ -118,7 +121,6 @@ def update_all(self, source_deployment_id, inter_deployment_dependencies):
          dependencies descriptions, but without a source_deployment(_id).
         :return: a list of created InterDeploymentDependencies IDs.
         """
-        # from celery.contrib import rdb; rdb.set_trace()  # noqa
```
tehasdf, line 121: haha this was probably me, but man, this has got to be ancient

mateumann:
Not needed anymore

PR #968, opened 2022-01-10
RD-3859 Add workflow_ctx interface for dep-update related calls

tehasdf:
Those are all the calls that are used by dep-update. Currently, the
workflow uses the rest-client directly for them. But I want it to
use the workflow-context, because then, we can also (later) implement
a local-handler for it, and test/use dep-update in local workflows
as well.

This PR is mostly very boring, because all it does, is exposes all
the missing stuf on workflow_ctx.

### DIFF
--- a/cloudify/workflows/workflow_context.py
+++ b/cloudify/workflows/workflow_context.py
@@ -935,6 +935,70 @@ class _WorkflowContextBase(object):
             force=force,
         )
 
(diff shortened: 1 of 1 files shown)


tehasdf:
note that this implements all the methods for the remote handler (which is required to keep the normal, remote, on-manager, dep-update working), but not on the local handler yet

geokala (approved):

PR #969, opened 2022-01-12
RD-3859 Implement the dep-update-related calls for local workflows

tehasdf:
This really continues #968, but all the boilerplate for the local
side this time.


tehasdf:
This needs more polishing as well, but I'm going to work on the local workflows in an iterative way

geokala (approved):

PR #971, opened 2022-01-18
RD-4003 Reverse the condition in wait_after_fail

tehasdf:
It was actually the wrong way around. A bit of refactor, too.


tehasdf:
the removed test is replaced by the inte-test added in cloudify-cosmo/cloudify-manager#3477

glukhman (approved):

PR #974, opened 2022-01-24
RD-4068 Store relationship drift with other instance's id

tehasdf:
There can be many relationships, so they have to be keyed by
something, and spec suggests for it to be just keyed by the other
instance's id. (other = the target or the source)

### DIFF
--- a/cloudify/plugins/workflows.py
+++ b/cloudify/plugins/workflows.py
@@ -863,7 +863,7 @@ def _make_check_drift_graph(
             graph.add_dependency(source_task, task)
             source_task.info = {
                 'instance_id': instance.id,
-                'prefix': 'source_relationships_'
+                'relationship_target': rel.target_id,
             }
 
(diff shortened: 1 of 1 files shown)


tehasdf:
eg 
```
Instance system properties:
	configuration_drift:                          {'ok': True, 'timestamp': '2022-01-24T12:39:16.475456', 'task': 'script_runner.tasks.run', 'result': {'a': 'b'}}
	target_relationships_configuration_drift:     {'y_r4fim9': {'ok': True, 'timestamp': '2022-01-24T12:39:18.961442', 'task': 'plug1.op2', 'result': {'a': 'd'}}, 'y_irwfds': {'ok': True, 'timestamp': '2022-01-24T12:39:19.019963', 'task': 'plug1.op2', 'result': {'a': 'd'}}}
```

mateumann (approved):

PR #985, opened 2022-02-06
revert pyyaml pin to 5.4.1

glukhman:


mateumann:
6.0 is too much of a change.  It is not bc compatible hence `yaml.load` https://github.com/cloudify-cosmo/cloudify-manager/blob/05a768667f185afeef4d96db2f244493925ee806/rest-service/manager_rest/upload_manager.py#L896 won't work.

mateumann:
Actually, let's try to make 6.0 work with our code… → https://github.com/cloudify-cosmo/cloudify-manager/pull/3511

geokala:
Closing this as it's more than 6 months old. Please reopen if we still want it.

PR #986, opened 2022-02-07
RD-3973 Stick to PyYAML==5.4.1

mateumann:
... because 6.0 is not back compatible making `yaml.load` fail e.g.
here: https://github.com/cloudify-cosmo/cloudify-manager/blob/05a768667f185afeef4d96db2f244493925ee806/rest-service/manager_rest/upload_manager.py#L896

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,14 +32,12 @@ if sys.version_info[:3] < (2, 7, 9):
         'requests==2.19.1',
         'fasteners==0.16.3',
     ]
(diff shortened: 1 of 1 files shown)


geokala (approved):

mateumann:
Actually, let's try to make 6.0 work with our code… → https://github.com/cloudify-cosmo/cloudify-manager/pull/3511

PR #991, opened 2022-02-15
RD-4115, RD-4116, RD-4282 `deployment_id` data type

mateumann:
The new data type `deployment_id` refers to deployments existing in the
system.  A valid input of that type is going to guarantee that given
deployment exists (possibly for a different tenant though).

Moreover a new `filter` constraints is added which might enforce
deployments to be associated with given tenants, have specific labels
or have display_names follow some pattern.

### DIFF
--- a/dsl_parser/constraints.py
+++ b/dsl_parser/constraints.py
@@ -21,7 +21,8 @@ from dsl_parser import functions, exceptions
 from dsl_parser.constants import (
     INPUTS,
     DEFAULT,
-    CONSTRAINTS as CONSTRAINT_CONST
+    CONSTRAINTS as CONSTRAINT_CONST,
+    TYPE,
 )
 
 _NOT_COMPARABLE_ERROR_MSG = "Value is not comparable, the Constraint " \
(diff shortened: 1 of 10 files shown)


dsl_parser/constraints.py
```diff
@@ -21,7 +21,8 @@
 from dsl_parser.constants import (
     INPUTS,
     DEFAULT,
-    CONSTRAINTS as CONSTRAINT_CONST
+    CONSTRAINTS as CONSTRAINT_CONST,
+    TYPE
```
geokala, line 25: A trailing comma here will mean the next time we add something to this it only changes one line.

dsl_parser/constraints.py
```diff
@@ -218,14 +222,94 @@ class Pattern(Constraint):
...
+                    "Filter '{0}' is not defined for data type 'deployment_id'"
+                    .format(name))
+        if len(deployments) < 1:
+            raise exceptions.ConstraintException(
+                "No matching deployment has been found for these constraints: "
+                "'{0}'".format(self.args))
+        return True
+
+    @staticmethod
+    def validate_deployment_labels(labels, deployments):
```
geokala, line 278: Can't this and the tenant thing be done as part of the get method by applying filters so you then just check the result count?

mateumann: No, not really.  I don't believe get method supports combining tenants, labels and display_name filters.

geokala: A func.partial could do it (if I'm thinking of the right thing), and then we should be making the database do the heavy lifting rather than doing it in python I think- not a big issue either way for small environments but could get painful in larger.

mateumann: Unfortunately (?) it's not going to be a simple query to the DB.  Most of the times we'll be using rest-client (see https://github.com/cloudify-cosmo/cloudify-manager/pull/3530/files#diff-a441270002aa034fc667e2efbfa3514f40909d712120a8fa3b3ecd8822193dbcR111)

geokala: Fair enough, then let's go with this and optimise later if we need to.

geokala (dismissed):

tehasdf (approved):

PR #1006, opened 2022-03-03
REST-client: ni.update() - keep back-compat wrt. args ordering

tehasdf:
In #954, we added `system_properties` to right after `runtime_properties`,
but that breaks compat with usage that passes all arguments positionally.

Instead, let's put it at the end.

### DIFF
--- a/cloudify_rest_client/node_instances.py
+++ b/cloudify_rest_client/node_instances.py
@@ -153,10 +153,10 @@ class NodeInstancesClient(object):
                node_instance_id,
                state=None,
                runtime_properties=None,
-               system_properties=None,
                version=1,
                force=False,
-               relationships=None):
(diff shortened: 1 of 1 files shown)


tehasdf:
internal usage does pass system_properties by name:
https://github.com/cloudify-cosmo/cloudify-common/blob/fa8453f7525dceff1479a522a8bd12d89cb5ff71/cloudify/plugins/workflows.py#L710
https://github.com/cloudify-cosmo/cloudify-common/blob/fa8453f7525dceff1479a522a8bd12d89cb5ff71/cloudify/plugins/workflows.py#L710
(to workflow_context, which then passes `**kwargs` anyway)

geokala (approved):

PR #1013, opened 2022-03-16
RD-4135 Define `secret_key` data type

mateumann:


dsl_parser/constants.py
```diff
@@ -97,7 +97,8 @@
 
 USER_PRIMITIVE_TYPES = ['string', 'integer', 'float', 'boolean', 'list',
                         'dict', 'regex', 'secret', 'textarea',
-                        'deployment_id', 'blueprint_id', 'capability_value']
+                        'deployment_id', 'blueprint_id', 'secret_key',
+                        'capability_value']
```
geokala, line 101: This does change the index of capability_value- do we care about the positions of types in this list?

mateumann: not at all

geokala: Excellent, I'll green then.

geokala (approved):

PR #1025, opened 2022-03-24
RD-4502 Hide useless tracebacks (incl. DSLParsingExc)

tehasdf:
Hmm. We could do even more work with making tracebacks nicer.
But for now, see these commits.

### DIFF
--- a/dsl_parser/exceptions.py
+++ b/dsl_parser/exceptions.py
@@ -106,6 +106,8 @@ class UnknownSysPropertyError(Exception):
 
 
 class DSLParsingException(Exception):
+    hide_traceback = True
+
     def __init__(self, err_code, *args):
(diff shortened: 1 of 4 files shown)


dsl_parser/exceptions.py
```diff
@@ -106,6 +106,8 @@ class UnknownSysPropertyError(Exception):
 
 
 class DSLParsingException(Exception):
+    hide_traceback = True
```
mateumann, line 109: From someone who likes grepping through the logs a lot: this sounds rather arbitrary, maybe we could make this parameter configurable (one day)?

tehasdf: OK, now we do include the full traceback in the log. Just not in `execution.error` (as returned by rest-client)

mateumann (approved):

PR #1030, opened 2022-03-29
RD-4280 Combine arguments of recursive functions

mateumann:
There are cases, where one attribute refers another, for example:

```
node_templates:
  test_tmp:
    type: tmp
    properties:
      a: [1,2,3,4]
      b: {get_attribute: [test_tmp, a]}
```

In this case getting n-th node of `b` property should give the same
result as getting the n-th node of `a` property.  In other words these
two should be equal:

```
capabilities:
  a0:
    value: {get_attribute: [test_tmp, a, 0]}
  b0:
    value: {get_attribute: [test_tmp, b, 0]}
```

This patch makes DSL parser work exactly like that by merging
_remaining_ parameters with the ones passed to intrinsic functions.
This way for `b0` we'll be truly getting value of
`{get_attribute: [test_tmp, a, 0]}` instead of
`[{get_attribute: [test_tmp, a]}, 0]`.

### DIFF
--- a/dsl_parser/functions.py
+++ b/dsl_parser/functions.py
@@ -645,7 +645,26 @@ def _get_attribute_from_node_instance(ni, node, attribute_path, path, fn_name):
                                     path,
                                     raise_if_not_found=False,
                                     func_name=fn_name)
+        if isinstance(value, list):
+            value = _merge_function_args(value)
+
     return value
+
+
+def _merge_function_args(items):
+    for index, item in enumerate(items):
+        if is_function(item):
+            break
+    else:
+        return items
+    item = items[index]
+    f_name = list(item.keys())[0]
+    f_args = item[f_name]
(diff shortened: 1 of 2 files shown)


dsl_parser/functions.py
```diff
@@ -645,9 +645,28 @@ def _get_attribute_from_node_instance(ni, node, attribute_path, path, fn_name):
...
+            value = _merge_function_args(value)
+
     return value
 
 
+def _merge_function_args(items):
+    for index, item in enumerate(items):
+        if is_function(item):
+            break
+    else:
```
geokala, line 658: I really don't like the use of for/else here.
Also, doesn't this way of writing it result in us not processing more than one function?

mateumann:
That's not the proper approach to deal with this issue.

PR #1052, opened 2022-05-13
RD-3788 typecheck operations inputs

mateumann:


dsl_parser/functions.py
```diff
@@ -122,10 +125,15 @@ def _register_entry_point_functions():
...
+    # {function_name: args, 'type': type_name} so let's look for
     # dicts of length 1 where the only key was registered as a function
-    return isinstance(value, dict) and len(value) == 1 \
-        and list(value.keys())[0] in TEMPLATE_FUNCTIONS
+    if not isinstance(value, dict):
+        return False
+    value_copy = copy.copy(value)
+    value_copy.pop('type', None)
+    return len(value_copy) == 1 \
+        and list(value_copy.keys())[0] in TEMPLATE_FUNCTIONS
```
tehasdf, line 136: `is_function` is called A LOT, so let's optimize it: copying is going to be slow. I'm thinking something like this:
```python
def is_function(value):
    if not isinstance(value, dict):
        return False
    if len(value) != 1 and len(value) != 2:
        return False
    for k in value:
        if k == 'type' or k in TEMPLATE_FUNCTIONS:
            continue
        else:
            return False
    return True
```

also, in that one place, we do `is_function` and then fetches it; maybe we don't want an `is_function`, maybe we want a `get_function` instead: it could be similar to the above,
```python
def get_function(value):
    [...]
    func = None
    for k in value:
        if k == 'type':
            continue
        func = TEMPLATE_FUNCTIONS.get(k)
    return func
```
and then all call-sites of `if is_function()` can be replaced to `if get_function() is None`

tehasdf: or rather return both k and v for get_function, because `parse` would need both

dsl_parser/functions.py
```diff
@@ -234,20 +244,32 @@ def validate(self, plan):
...
 
         if is_function(self.input_value):
             raise exceptions.FunctionEvaluationError(
                 '{0}: found an unresolved argument: {1}'
                 .format(self.name, self.input_value))
-        return handler.get_input(self.input_value)
+        return_value = return_value or handler.get_input(self.input_value)
+
+        if self.return_type:
+            from dsl_parser.utils import parse_simple_type_value
```
tehasdf, line 264: hmm unfortunate. We can untangle this by moving is_function and TEMPLATE_FUNCTIONS into utils.py...?

mateumann: not really, already tried that

tehasdf: oh? I just moved them and it seems fine, `= 1051 passed, 1 warning in 9.26s =`

mateumann: You can try running it on the actual manager – I think that's where the problems began.  It was late on the other hand, I will check one more time

cloudify/workflows/workflow_context.py
```diff
@@ -672,6 +675,9 @@ def _execute_operation(self,
         final_kwargs = self._merge_dicts(merged_from=kwargs,
                                          merged_into=operation_properties,
                                          allow_override=allow_kwargs_override)
+        operation_properties_types = op_struct.get('inputs_types')
+        if operation_properties_types:
```
tehasdf, line 679: either `is not None` here, or handle None in `_validate_types` (currently it requires a dict, not optional[dict])

tehasdf (dismissed):

dsl_parser/utils.py
```diff
@@ -436,3 +438,23 @@ def add_values_node_description(data):
     for k, v in data.items():
         result[k] = {'values': v}
     return result
+
+
+def get_function(value):
+    """Get a template function if the value represents it"""
+    # functions use the syntax {function_name: args}, or
+    # {function_name: args, 'type': type_name} so let's look for
+    # dicts of length 1 where the only key was registered as a function
```
geokala, line 447: This comment is a lie.

dsl_parser/utils.py
```diff
@@ -436,3 +438,23 @@ def add_values_node_description(data):
...
+    if not isinstance(value, dict):
+        return None
+    if len(value) != 1 and len(value) != 2:
```
geokala, line 450: This might be better written as "if not 0 < len(value) < 3:"?

dsl_parser/utils.py
```diff
@@ -436,3 +438,24 @@ def add_values_node_description(data):
...
+
+def get_function(value):
+    """Get a template function if the value represents it"""
+    # functions use the syntax {function_name: args}, or
+    # {function_name: args, 'type': type_name} so let's look for
+    # dicts of length 1 where the only key was registered as a function
+    # or length of 2 where the other key is 'type'
+    if not isinstance(value, dict):
+        return None
+    if 0 < len(value) < 3:
```
geokala, line 451: I think this logic is inverted? This means we'll return no function if we have a length of 1 or 2 where we in fact want the opposite?

geokala (approved):

PR #1055, opened 2022-05-18
RD-4923 Add manager/agent name to ops & events

tehasdf:
Operations and events will now know which manager or agent created them.

### DIFF
--- a/cloudify/logs.py
+++ b/cloudify/logs.py
@@ -21,13 +21,17 @@ import datetime
 
 from cloudify import constants, manager
 from cloudify import event as _event
-from cloudify.utils import (get_execution_creator_username,
-                            get_execution_id,
-                            is_management_environment,
-                            ENV_AGENT_LOG_LEVEL,
-                            ENV_AGENT_LOG_DIR,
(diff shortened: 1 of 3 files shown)


cloudify/logs.py
```diff
@@ -371,9 +375,18 @@ def _publish_message(message, message_type, logger, skip_send=False):
...
+        events = None
     else:
-        client.events.create(events=[message], execution_id=execution_id)
+        logs = None
+        events = [message]
+    client.events.create(
+        events=events,
+        logs=logs,
+        execution_id=execution_id,
+        agent_name=get_daemon_name(),
```
geokala, line 387: Should this perhaps just be executor_name which does get_daemon_name() or get_manager_name()?

tehasdf: I kinda want to go with separate, because it's just a string, it's hard to tell from just looking at it whether it's a manager or an agent.
Additionally, I envision a future in which there could be BOTH set, when you have multiple instances of a mgmtworker running on a single host.

geokala: <thumbs up>

geokala (approved):

PR #1058, opened 2022-05-23
RD-4355 One file `cfy` executable

mateumann:
That might fix pyoxidizer issue with building cloudify-cli


mateumann:
instead → https://github.com/cloudify-cosmo/cloudify-common/pull/1074

PR #1082, opened 2022-06-09
Add a few magic methods to CloudifyWorkflowNode/Instance

tehasdf:
- the repr is just nice, duh
- NodeInstance is already put into sets, so it very much SHOULD have
  a hash+eq. It doesn't currently, which makes it break when some
  instances are from before a `wctx.refresh_node_instances()` and
  some are from after (then they compare not equal, which is... less
  than useful)

### DIFF
--- a/cloudify/workflows/workflow_context.py
+++ b/cloudify/workflows/workflow_context.py
@@ -235,6 +235,12 @@ class CloudifyWorkflowNodeInstance(object):
         self._logger = None
 
     def __eq__(self, other):
+        """Compare node instances
+
+        A node instance is always going to be equal with itself, even if it
+        was fetched multiple times, and it was mutated by the user.
+        This allows storing node instances in sets, and as dict keys.
+        """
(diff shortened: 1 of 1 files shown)


cloudify/workflows/workflow_context.py
```diff
@@ -234,6 +234,21 @@ def __init__(self, ctx, node, node_instance, nodes_and_instances):
 
         self._logger = None
 
+    def __eq__(self, other):
+        if not isinstance(other, self.__class__):
+            return False
+        return self.id == other.id
```
geokala, line 240: Shouldn't this take into account other attributes in case of things like dep modifications?

tehasdf: No, why? :)

tehasdf: I've said in a docstring what I think it should do. But anyway, this way around it's more useful

geokala (approved):

PR #1101, opened 2022-07-05
RD-4466 check_drift: add ignore_failure

tehasdf:
The check_drift graph is going to be called from update, where we'll
have a special way of handling failures (if check_drift for a node
fails, we'll reinstall the instance). So, no need to throw the
error up.

...and also do the same for check_status. It's nicer than the
try/except.

### DIFF
--- a/cloudify/plugins/workflows.py
+++ b/cloudify/plugins/workflows.py
@@ -213,14 +213,9 @@ def auto_heal_reinstall_node_subgraph(
             ctx,
             node_instance_ids=[ni.id for ni in sub_node_instances],
             name='check_status',
+            ignore_failure=True,
         )
-        try:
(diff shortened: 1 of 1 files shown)


cloudify/plugins/workflows.py
```diff
@@ -213,14 +213,9 @@ def auto_heal_reinstall_node_subgraph(
             ctx,
             node_instance_ids=[ni.id for ni in sub_node_instances],
             name='check_status',
+            ignore_failure=True,
         )
-        try:
```
tehasdf, line 217: this is tested by `test_builtin_workflows.py::TestHealOperation::test_failing_status_heal_operation`

geokala (approved):

PR #1112, opened 2022-07-11
RD-5229 Snapshot create improvements

glukhman:


cloudify_rest_client/snapshots.py
```diff
@@ -127,7 +128,8 @@ def create(self,
             'include_credentials': include_credentials,
             'include_logs': include_logs,
             'include_events': include_events,
-            'queue': queue
+            'queue': queue,
+            'tempdir_path': tempdir_path
```
geokala, line 132: If you put a comma after tempdir_path then the next time this is changed it won't require a two line change.

geokala (approved):

PR #1119, opened 2022-08-02
RD-5232 Print the server-side cause of `CloudifyClientError`

mateumann:
This way we get slightly more informative, but not too verbose, `CloudifyClientError`s

### DIFF
--- a/cloudify_rest_client/exceptions.py
+++ b/cloudify_rest_client/exceptions.py
@@ -28,10 +28,9 @@ class CloudifyClientError(Exception):
     def __str__(self):
         message = self._message
         if self.server_traceback:
-            traceback_list = \
-                [line for line in self.server_traceback.split('\n') if line]
-            if traceback_list:
-                message = '{0} ({1})'.format(self._message, traceback_list[-1])
+            _, _, last_line = self.server_traceback.strip().rpartition('\n')
+            if last_line:
+                message = '{0} ({1})'.format(self._message, last_line)
         if self.status_code != -1:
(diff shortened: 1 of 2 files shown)


mateumann:
Instead of:

`An error occurred on the server: 500: Internal error occurred in manager REST server - InternalServerError: 500 Internal Server Error: The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application.`

we would get:

`An error occurred on the server: 500: Internal error occurred in manager REST server - InternalServerError: 500 Internal Server Error: The server encountered an internal error and was unable to complete your request. Either the server is overloaded or there is an error in the application. (MemoryError)`

cloudify_rest_client/exceptions.py
```diff
@@ -26,9 +26,15 @@ def __init__(self, message, server_traceback=None,
         self._message = message
 
     def __str__(self):
+        message = self._message
+        if self.server_traceback:
+            traceback_list = \
+                [line for line in self.server_traceback.split('\n') if line]
```
tehasdf, line 32: we only care about the last line, so how about something like...

```
_, _, last_line = self.server_traceback.strip().rpartition('\n')
if last_line:
```

mateumann: not really, usually the last line would be just empty string

tehasdf: hence the `.strip()` ;)

tehasdf (approved):

PR #1121, opened 2022-08-04
RD-5524 Plugin resolver: use re.search, not re.match

tehasdf:
re.match matches at the beginning of the string! Attempting to match
something at the end is futile. That's what we have `re.search` for.


tehasdf:
This caused `manager_rest/test/infrastructure/test_resolver_with_catalog_support.py::TestPluginParseWithResolver::test_fetches_proper_plugin_yaml` to reliably fail locally for me.

Why not on jenkins? I'm not sure. Perhaps the `glob.glob` call gives out filenames depending on the filesystem, and on jenkins, `plugin.yaml` comes out before `plugin_1_3.yaml`, but on my machine, it's the opposite?

tehasdf:
See jira for on-manager impact

mateumann (approved):

PR #1122, opened 2022-08-12
RD-5599 Return the "latest" plugin*.yaml in case dsl_version is not provided

mateumann:
This fixes broken `manager_rest.test.infrastructure.test_resolver_with_catalog_support.TestPluginParseWithResolver testMethod=test_fetch_plugin_yaml_matching_dsl_version` test and also `Resolver.fetch_import` for plugins in general in case `dsl_version` is not provided.

This scenario (unknown `dsl_version`) should rather not happen in real-life scenario.

### DIFF
--- a/dsl_parser/import_resolver/default_import_resolver.py
+++ b/dsl_parser/import_resolver/default_import_resolver.py
@@ -179,3 +179,7 @@ class DefaultImportResolver(AbstractImportResolver):
         for yaml_file in yaml_files:
             if not re.search(r"_\d_\d+\.yaml$", yaml_file):
                 return [yaml_file]
+
+        # If dsl_version is not provided, return the last yaml_file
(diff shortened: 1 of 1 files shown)


geokala:
There should probably be a ticket associated with this, assuming it's the correct behaviour.

One good reason not to have it behave like this would be because then we'll accept any random yaml fail and not fail fast.

mateumann:
Yeah, so I was thinking, since it's really very uncommon to import a plugin.yaml and not know the dsl_version of the base blueprint, maybe instead let's just "fix" the test.

mateumann:
Actually let's do it like that: https://github.com/cloudify-cosmo/cloudify-manager/pull/3792

tehasdf (approved):

PR #1125, opened 2022-08-18
RD-5457 Drop Python2 compatibility

mateumann:


tehasdf (approved):

cloudify/workflows/local.py
```diff
@@ -121,7 +121,7 @@ def execute(self,
             return rv
         except Exception as e:
             self.storage.execution_ended(execution_id, e)
-            reraise(e.__class__, e, sys.exc_info()[2])
+            raise e.with_traceback(sys.exc_info()[2])
```
tehasdf, line 124: this one doesn't make much sense, just reraising the same exception we're handling
```suggestion
            raise
```

tehasdf (approved):

PR #1136, opened 2022-09-06
RD-5243 Keep trying to update operation's state

mateumann:
Someone might disconnect an Ethernet cable, or update Windows drivers
for the NIC, or Ethernet switch might be restarted, or ...  So instead
of failing, let's keep trying to contact manager to update the
operation's state.


cloudify/utils.py
```diff
@@ -1058,3 +1060,17 @@ def uuid4():
         uuid_as_hex[16:20],
         uuid_as_hex[20:]
     )
+
+
+def keep_trying_http(func, total_timeout_sec=None):
```
geokala, line 1065: Should we default to no timeout?

mateumann: I think we should, at least for agents.  They probably shouldn't be allowed to perform any new operations until the current one is finished and successfully reported as such.
Of course, your question is more interesting in a broader scope.  Maybe we would wrap some more requests (not just `CloudifyContext.update_operation()`) with `keep_trying_http`.  I'll take a look.

geokala: Not performing new operations probably makes sense (will they anyway?), but they should probably be able to error out at some point.

Also, I note there's no backoff here, so if the server is being overwhelmed by traffic then they'll just hammer it as fast as they can.
There's also no logging so one can't tell when a problem is occurring by looking anywhere.

cloudify/utils.py
```diff
@@ -1058,3 +1062,19 @@ def uuid4():
...
+    def wrapper(func, *args, **kwargs):
+        timeout_at = None if total_timeout_sec is None \
+            else datetime.utcnow() + timedelta(seconds=total_timeout_sec)
+        while True:
+            try:
+                return func(*args, **kwargs)
+            except (ConnectionError, Timeout):
+                if timeout_at and datetime.utcnow() > timeout_at:
+                    raise
+            time.sleep(random.randint(1, max_delay_sec))
```
geokala, line 1078: Some form of logging would probably be good.

cloudify/utils.py
```diff
@@ -1058,3 +1062,23 @@ def uuid4():
...
+        timeout_at = None if total_timeout_sec is None \
+            else datetime.utcnow() + timedelta(seconds=total_timeout_sec)
+        while True:
+            try:
+                return func
+            except (ConnectionError, Timeout):
+                if timeout_at and datetime.utcnow() > timeout_at:
+                    raise
+            delay = random.randint(1, max_delay_sec)
+            logger.info(f'Will retry {func} in {delay} seconds')
```
geokala, line 1081: This should probably include the error so that you can see what it is (and maybe be able to fix it or figure out why it happened)

mateumann: Oh, right.  I was expecting client to produce some logs but as it turns out, it's the case only if debugging is enabled…  Thanks.

tehasdf (dismissed):

cloudify/utils.py
```diff
@@ -1058,3 +1063,53 @@ def uuid4():
...
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            timeout_at = None if total_timeout_sec is None \
+                else datetime.utcnow() + timedelta(seconds=total_timeout_sec)
+            while True:
+                try:
+                    return func(*args, **kwargs)
+                except (ConnectionError, Timeout) as ex:
+                    if timeout_at and datetime.utcnow() > timeout_at:
+                        logger.info(f'Finished retrying {func}: '
```
geokala, line 1101: logger.error

cloudify/utils.py
```diff
@@ -1058,3 +1063,53 @@ def uuid4():
...
+                                    f'total timeout of {total_timeout_sec} '
+                                    f'seconds exceeded: {ex_to_str(ex)}')
+                        raise
+                    delay = random.randint(0, max_delay_sec)
+                    logger.info(f'Will retry {func} in {delay} seconds: ' +
```
geokala, line 1106: This should probably be either warn or debug, not info. I'm not sure which- depends on how often we're expecting this to be and whether we consider it to be something that is at all remarkable.
I'm leaning towards warn.

cloudify/utils.py
```diff
@@ -1058,3 +1063,53 @@ def uuid4():
...
+                                ex_to_str(ex))
+                    time.sleep(delay)
+                except Exception as ex:
+                    logger.info(f'Will not retry {func}: the encountered '
```
geokala, line 1110: error

You might even be able to do... is it logger.exception here?

cloudify/utils.py
geokala: (or .exception)

cloudify/utils.py
```diff
@@ -1058,3 +1063,53 @@ def uuid4():
...
+                                         f'seconds exceeded: {ex_to_str(ex)}')
+                        raise
+                    delay = random.randint(0, max_delay_sec)
+                    logger.warning(f'Will retry {func} in {delay} seconds: ' +
+                                   ex_to_str(ex))
+                    time.sleep(delay)
+                except Exception as ex:
+                    logger.exception(f'Will not retry {func}: the encountered '
+                                     'error cannot be fixed by retrying: ' +
+                                     ex_to_str(ex))
```
geokala, line 1112: With logging.exception, you don't need to provide the exception to it, you just need to use it in an exception handler and it'll put the traceback at the end after a newline:
`>>> def testit():
...     try:
...         raise ValueError('test')
...     except Exception:
...         logger.exception('There was a bad.')
... 
>>> testit()
There was a bad.
Traceback (most recent call last):
  File "<stdin>", line 3, in testit
ValueError: test
`

geokala: After some external discussion: These should probably be logger.error anyway, the full traceback is almost certainly just noise given the error types we have.

tehasdf (approved):

PR #1145, opened 2022-09-12
RD-5865 get REST protocol and token from env

adrian-polanczyk-codilime:


cloudify/utils.py
```diff
@@ -223,6 +223,15 @@ def get_manager_file_server_root():
     return os.environ[constants.MANAGER_FILE_SERVER_ROOT_KEY]
 
 
+def get_manager_rest_service_protocol():
+    protocol = os.environ[constants.REST_PROTOCOL_KEY]
+
+    if protocol is None:
+        protocol = constants.SECURED_PROTOCOL
+
+    return protocol
```
tehasdf, line 232: I note that `os.environ[nonexistent]` throws a KeyError, so the `is None` check is futile.
This perhaps just wants to be:
```suggestion
    return os.environ.get(
        constants.REST_PROTOCOL_KEY,
        constants.SECURED_PROTOCOL,
    )
```

(same for the other func)

tehasdf (commented):
Cool but the PR title doesn't need to be this long:
`RD-5865 - Microservice - API Admin Token and Protocol provided by env…`
as you note it doesn't even fit on github :)

It doesn't really need to mention "microservices" at all, the change can very well stand on its own. Eg. just `RD-5865 get REST protocol and token from env` would be enough.
If you want, the PR description can mention "micro"services.

tehasdf (approved):

PR #1150, opened 2022-09-14
Move mock_rest_client to cloudify.test_utils

tehasdf:
Other packages - notably, cloudify-agent, want to use that mock in their tests, however in here (setup.py) we say
`exclude=('cloudify.tests*',`.

And so, when installing common in cloudify-agent tests, those tests don't get that mock.

Move it to cloudify.test_utils, so that cloudify-agent tests will get it.

Also remove unused mock_module.


tehasdf:
I don't know why it used to work before, presumably pip and setuptools didn't honor the `exclude`?

geokala (approved):

PR #1152, opened 2022-09-19
RD-5982 Use `distro` instead of `platform` module

mateumann:


cloudify/plugin_installer.py
```diff
@@ -456,7 +462,7 @@ def _is_plugin_supported(plugin):
 
 def _platform_and_distro():
     current_platform = wagon.get_platform()
-    distribution, _, distribution_release = platform.linux_distribution(
+    distribution, _, distribution_release = linux_distribution(
```
tehasdf, line 465: `linux_distribution` can be None (re importerror), so it'd be good if this was handled here. Just a `if is None: raise NonRecoverable("distro must be installed")` would be OK

tehasdf (approved):

PR #1154, opened 2022-09-21
RD-5832 Imported plugin properties

mateumann:


dsl_parser/elements/imports.py
```diff
@@ -101,10 +137,17 @@ def validate(self, **kwargs):
         imports = [i.value for i in self.children()]
         imports_set = set()
         for _import in imports:
-            if _import in imports_set:
-                raise exceptions.DSLParsingFormatException(
-                    2, 'Duplicate imports')
-            imports_set.add(_import)
+            if isinstance(_import, str):
```
geokala, line 140: If you make the conditional set import_key = _import (or import_key = _validate_import_dict(_import) when a dict), then you can write the raising/adding just once.

dsl_parser/tests/test_plugin_properties.py
```diff
@@ -0,0 +1,185 @@
...
+            self.parse_1_5(yaml)
+        assert p.call_count == 1
+
+    def test_valid(self):
+        yaml = """
+imports:
+  - cloudify/types/types.yaml
+  - plugin:dummy:
+      aws_access_key_id: foobar
+      aws_secret_access_key: {get_secret: my_secret}
```
geokala, line 113: Shouldn't these tests also be asserting on the properties they see for the plugin? I'm not currently seeing any where we assert that we get the properties we expect (unless those properties are an empty dict)?

dsl_parser/tests/test_plugin_properties.py
```diff
@@ -0,0 +1,185 @@
...
+
+node_templates:
+  dummy_node:
+    type: type
+    interfaces:
+      dummy_install:
+        install: dummy.install
+"""
+        plan = self.parse_1_5(yaml)
+        assert plan['deployment_plugins_to_install'][0]['properties'] == {}
```
geokala, line 79: Should this really be an empty dict? I would've expected it should either be an error (if we require the properties) or a dict containing all the possible properties with them set to None?

dsl_parser/tests/test_plugin_properties.py
mateumann: The other test https://github.com/cloudify-cosmo/cloudify-common/blob/447564e2385a8a382f5ee76504d835a577095a29/dsl_parser/tests/test_plugin_properties.py#L52-L57 does that.  This one is basically about the syntax,

dsl_parser/tests/test_plugin_properties.py
mateumann: Yes, the plugin hasn't got any properties, so an empty dict should be appropriate to represent that

dsl_parser/tests/test_plugin_properties.py
geokala: That's only asserting that they're strings. It's not asserting that they have the actual content they're assigned.

dsl_parser/tests/test_plugin_properties.py
geokala: Doesn't the plugin have three aws_<thing> properties defined?

dsl_parser/tests/test_plugin_properties.py
mateumann: no, it asserts that all properties are of type `string` (which is most interesting), but sure, I could add a few more asserts

dsl_parser/tests/test_plugin_properties.py
mateumann: No, it does not, we're testing this yaml in this case:
```yaml
plugins:
  dummy:
    executor: central_deployment_agent
    package_name: testing
    source: dummy
    properties_description: Empty credentials

node_types:
  type: {}

node_templates:
  dummy_node:
    type: type
    interfaces:
      dummy_install:
        install: dummy.install
```

geokala (approved):

PR #1162, opened 2022-10-26
RD-6168 Support py36 plugins

mateumann:


geokala (approved):

cloudify/plugin_installer.py
```diff
@@ -558,3 +558,16 @@ def get_pth_dir(venv=None):
...
         raise NonRecoverableError('Unsupported OS: {0}'.format(os.name))
+
+
+def _python_executable(plugin):
+    if 'supported_py_versions' not in plugin:
+        return sys.executable
+    if 'py310' in plugin['supported_py_versions']:
+        return sys.executable
+    if 'py36' in plugin['supported_py_versions']:
+        return '/usr/bin/python3.6'
```
tehasdf, line 569: This file won't always exist, so maybe let's have a `os.path.exists` or something like that

tehasdf: (of course, don't NonRecoverable eagerly - maybe 3.6 doesn't exist, but there are other supported_py_versions for that plugin that do exist)

cloudify/plugin_installer.py
```diff
@@ -558,3 +558,16 @@ def get_pth_dir(venv=None):
         return '{0}/lib/python{1}/site-packages'.format(prefix, version)
     else:
         raise NonRecoverableError('Unsupported OS: {0}'.format(os.name))
+
+
+def _python_executable(plugin):
+    if 'supported_py_versions' not in plugin:
+        return sys.executable
+    if 'py310' in plugin['supported_py_versions']:
+        return sys.executable
```
tehasdf, line 567: Let's also have a check here, that 3.10 is indeed the version we're on (`sys.version_info`)

We'll soon probably add 3.11 here too!

PR #1170, opened 2022-11-03
RD-6241 yamllint on blueprint_upload workflow

mateumann:


tehasdf (approved):
I do expect we'll have to go back on the indentation error but for now why not :P

tehasdf (approved):

PR #1177, opened 2022-11-09
RD-6264 Create and list Secret Providers

adrian-polanczyk-codilime:


cloudify_rest_client/secret_provider.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class SecretProviderClient(object):
+    def __init__(self, api):
+        self.api = api
+
+    def create(
+            self,
+            secret_provider_name,
+            secret_provider_type,
```
tehasdf, line 46: maybe just `name` and `type`?

cloudify_rest_client/secret_provider.py
```diff
@@ -0,0 +1,68 @@
...
+            connection_parameters,
+            tenant_name,
+            visibility=VisibilityState.TENANT,
+    ):
```
tehasdf, line 50: for sure needs a docstring, describing the params

cloudify_rest_client/secret_provider.py
```diff
@@ -0,0 +1,68 @@
...
+        data = {
+            'name': secret_provider_name,
+            'type': secret_provider_type,
+            'connection_parameters': connection_parameters,
+            'tenant_name': tenant_name,
+            'visibility': visibility,
+        }
+        response = self.api.put('/secret-provider', data=data)
```
tehasdf, line 58: the url should be plural too probably. Everything else is :P

cloudify_rest_client/secret_provider.py
```diff
@@ -0,0 +1,68 @@
...
+    def name(self):
+        return self.get('name')
+
+    @property
+    def type(self):
+        return self.get('type')
+
+    @property
+    def connection_parameters(self):
+        return self.get('connection_parameters')
```
tehasdf, line 20: at least those should probably have a docstring

cloudify_rest_client/secret_provider.py
```diff
@@ -0,0 +1,68 @@
+from cloudify_rest_client.responses import ListResponse
+from cloudify_rest_client.constants import VisibilityState
+
+
+class SecretProvider(dict):
+    def __init__(self, secret):
+        super(SecretProvider, self).__init__()
+        self.update(secret)
```
tehasdf, line 8:
```suggestion
    def __init__(self, provider):
        super(SecretProvider, self).__init__()
        self.update(provider)
```

cloudify_rest_client/secret_provider.py
```diff
@@ -0,0 +1,68 @@
...
+class SecretProviderClient(object):
+    def __init__(self, api):
+        self.api = api
+
+    def create(
+            self,
+            secret_provider_name,
+            secret_provider_type,
+            connection_parameters,
+            tenant_name,
```
tehasdf, line 48:
```suggestion
            tenant_name=None,
```
this must not be required, but needs to be defaulted (from the client's header)

tehasdf: actually nope, just remove `tenant_name` from here entirely (and don't pass it through from the CLI).

The restclient object, when it's created, already has the selected tenant name in its headers

tehasdf (approved):

PR #1180, opened 2022-11-10
RD-6279 upgrade to networkx ==2.8.8

mateumann:


setup.py
```diff
@@ -37,10 +37,10 @@
     ],
     extras_require={
         # for running workflows (in the mgmtworker and the cli), as opposed
-        # to eg. just executing operations (in the agent)
+        # to e.g. just executing operations (in the agent)
         'dispatcher': [
             'PyYAML==6.0',
-            'networkx==1.11',
+            'networkx==2.8.8',
```
tehasdf, line 43: how about `networkx>2,<3` or something, let's avoid additional strict version pins in setup.py

(will also require rebuilding reqs.txt on the cloudify-manager side, of course)

tehasdf (approved):

PR #1181, opened 2022-11-14
RD-6061 Provide get, update and delete commands for Secrets Providers

adrian-polanczyk-codilime:


cloudify_rest_client/secrets_providers.py
```diff
@@ -100,6 +105,46 @@ def create(
...
 
+    def update(
+            self,
+            name,
+            _type,
+            connection_parameters,
+            visibility=VisibilityState.TENANT,
+    ):
+        """
+        Update already exists Secrets Provider.
```
tehasdf, line 116:
```suggestion
        Update an existing Secrets Provider.
```
just grammar

cloudify_rest_client/secrets_providers.py
```diff
@@ -100,6 +105,46 @@ def create(
...
+
+        :param name: Secrets Provider name
+        :type name: str
+        :param _type: Secrets Provider type
+        :type _type: str
+        :param connection_parameters: Secrets Provider connection parameters
+        :type connection_parameters: dict
+        :param visibility: The visibility of the secret, can be 'private',
```
tehasdf, line 124:
```suggestion
        :param visibility: The visibility of the Secrets Provider, can be 'private',
```

cloudify_rest_client/secrets_providers.py
```diff
@@ -100,6 +105,46 @@ def create(
...
+                           'tenant' or 'global'
+        :type visibility: str
+        :returns: New secrets provider metadata
+        :rtype: Dict[str]
```
tehasdf, line 128: `Dict[str]` isn't a type, `Dict[str, Any]` would be a type, but really, I think we can just say...
```suggestion
        :rtype: SecretsProvider
```

cloudify_rest_client/secrets_providers.py
```diff
@@ -100,6 +105,46 @@ def create(
...
+        }
+        data = dict((k, v) for k, v in data.items() if v is not None)
+
+        response = self.api.patch(f'/secrets-providers/{name}', data=data)
+
+        return SecretsProvider(response)
+
+    def delete(self, name):
+        """
+        Deletes single Secrets Providers.
```
tehasdf, line 144:
```suggestion
        Delete a Secrets Provider.
```
tbh I'm not sure if we use the `Deletes` for or the `Delete` form. But we do have `Update` here, so might as well do `Delete` as well.

PR #1182, opened 2022-11-16
RD-6368 Add secret schema to secret create

glukhman:


cloudify_rest_client/secrets.py
```diff
@@ -89,15 +93,26 @@ def create(self,
...
         :rtype: Dict[str]
 
         """
+        if schema and schema.get('type') != 'string':
+            try:
+                value = json.loads(value)
+            except json.decoder.JSONDecodeError as e:
+                raise CloudifyClientError(
+                    f'Error decoding JSON string "{value}" as '
+                    f'type "{schema.get("type")}": {e.args[0]}')
```
tehasdf, line 108: maybe no need for this validation. Server-side is enough? Especially as it's not correct :P
eg. `.create(value=1, schema={'type': 'number'})`

glukhman: We need to handle `json.loads(value)` failures gracefully somehow. I can add a TypeError here

tehasdf:
plz see comment under cloudify-cosmo/cloudify-manager#3924 as well

tehasdf (approved):

PR #1186, opened 2022-11-22
RD-5953 Return context with IDDs

tehasdf:
When computing IDDs, it's not enough to just return the function representation, because to evaluate that function later, we also need the context!

That is because e.g. `{get_capability: [dep1, {get_attribute: [SELF, x]}]}` depends on SELF, so we need to know the node instance id in order to evaluate that.

So, in this PR, we make sure to return IDDs in a new format: instead of the old:
```python
{
   'nodes.n1.properties.xxx': {'get_capability': ['dep1', 'x']},
   'nodes.n2.properties.yyy': {'get_capability': ['dep2', 'y']}
}
```
we will now be returning IDDs in the deployment plan in the format of:
```python
[
    {
        'function_identifier': 'nodes.n1.properties.xxx',
        'target_deployment': {'get_capability': ['dep1', 'x']},
        'context': {
            'SELF': 'n1_abcdef',
        }
    }
]
```

In order to do this,
1. Rewrite IDD gathering - instead of registering to plan while evaluating functions, have a separate explicit step that goes through the plan and finds the functions
2. After creating the deployment plan, correlate the IDDCreatingFunctions, with actual node instances, so if there's a IDD function on `nodes.n1`, and there's 10 node-instances of n1, there will be 10 instances of IDD returned
3. ...and a bit unrelated, but since in 1 we learned how to do this, apply this plan-traversing idea to secrets as well, so we can lose the old `global` approach

(I do recommend viewing commits separately)

### DIFF
--- a/dsl_parser/tasks.py
+++ b/dsl_parser/tasks.py
@@ -168,8 +168,8 @@ def _find_rel_target_by_iddspec(relationships, idd_spec):
     """
     assert idd_spec.scope == 'node_template_relationship'
 
-    target_name = idd_spec.context['target']  # target _node_ name
-    rel_type = idd_spec.context['type']
+    target_name = idd_spec.context['target_node_name']
+    rel_type = idd_spec.context['relationship_type']
 
     for relationship in relationships:
         if (
@@ -215,7 +215,7 @@ def _format_idds(plan, dep_specs):
         if idd_spec.scope == 'node_template':
             # the IDDSpec is a node one (e.g. in node properties):
             # create an instance of IDD, for every node-instance of the node
-            node_name = idd_spec.context['node']
+            node_name = idd_spec.context['node_name']
             for ni in nis_by_node[node_name]:
                 idd = idd_base.copy()
                 idd['context'] = {
@@ -227,7 +227,7 @@ def _format_idds(plan, dep_specs):
             # the IDDSpec is a relationship one (e.g. in relationship operation
             # inputs) - create an instance of IDD for every node-instance of
             # the source node
-            node_name = idd_spec.context['node']
+            node_name = idd_spec.context['source_node_name']
 
             for ni in nis_by_node[node_name]:
                 target_id = _find_rel_target_by_iddspec(
--- a/dsl_parser/tasks.py
+++ b/dsl_parser/tasks.py
@@ -155,6 +155,100 @@ def _find_idd_creating_functions(plan) -> List[functions.IDDSpec]:
     return handler.dependencies
 
 
+def _find_rel_target_by_iddspec(relationships, idd_spec):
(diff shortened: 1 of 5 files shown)


dsl_parser/tasks.py
```diff
@@ -155,6 +155,100 @@ def _find_idd_creating_functions(plan) -> List[functions.IDDSpec]:
...
+    """Find the target instance that matches the relationship in IDDSpec
+
+    The node-instance can have many relationships, so we need to find one
+    that matches the target node name & type.
+    If there's multiple relationships of the same type to the same node,
+    then we will just select the first instance, because we have no way
+    to distinguish them.
+    (that case shouldn't be possible anyway)
+    """
+    assert idd_spec.scope == 'node_template_relationship'
```
geokala, line 169: Can we either convert this to a check that won't disappear with -O, or just drop it if we don't care whether it's executed?

dsl_parser/tasks.py
```diff
@@ -155,6 +155,100 @@ def _find_idd_creating_functions(plan) -> List[functions.IDDSpec]:
...
+
+    target_name = idd_spec.context['target']  # target _node_ name
```
geokala, line 171: Any reason to comment it rather than just calling it target_node_name?

dsl_parser/tasks.py
tehasdf: no, this is for the human reading this (or writing this)
I prefer to write it like this, rather than in the docstring

dsl_parser/tasks.py
```diff
@@ -155,6 +155,100 @@ def _find_idd_creating_functions(plan) -> List[functions.IDDSpec]:
...
+        - function_identifier - the IDDCreatingFunction path
+        - target_deployment - the deployment id, or a dict representing
+          a function call
+        - context - a dict containing possibly the keys SELF, TARGET, SOURCE -
+          this is the context that can be passed into function evaluation,
+          when evaluating the target_deployment (if it is a function)
+    """
+    idds = []
+    nis_by_node = {}
+    for ni in plan.get('node_instances', []):
```
geokala, line 205: Not worth doing as part of this, but I wonder if in general it would make sense to make the plan generate node instances as a dict with lists of node instances keyed on node_id?

geokala: (because I suspect we need to do something like this in several places?)

dsl_parser/tasks.py
geokala: OK, always strikes me as pointless but fine.

dsl_parser/tasks.py
geokala: OK- let's put it on the "one day, maybe today- maybe not" pile.

geokala (approved):

PR #1205, opened 2022-12-13
RD-6064 Provider options for Secret resource

adrian-polanczyk-codilime:


cloudify_rest_client/secrets.py
```diff
@@ -109,6 +119,8 @@ def create(self,
         :type schema: dict
         :param provider: A Secrets Provider name
         :type provider: str
+        :param provider_options: A JSON Secrets Provider options
```
tehasdf, line 122: nothing JSON about it, it's just a dict

it will be encoded as json later yes

cloudify_rest_client/secrets.py
```diff
@@ -75,20 +75,30 @@ def provider_name(self):
         """
         return self.get('provider_name')
 
+    @property
+    def provider_options(self):
+        """
+        :return: the secret's provider options
```
tehasdf, line 81: plz more info about this, like an example or something. "provider options" is just not clear enough

tehasdf (approved):

PR #1219, opened 2023-01-04
RD-6652 Scale: don't run unnecessary operations

tehasdf:
When rolling back after a failed scale, only run the operations that are actually needed to roll back.

For example, if an instance failed creating, we don't need to `stop` it, only `delete`.


tehasdf:
This can't be tested here currently, because local workflows don't implement deployment modifications. But I'm adding a test in cloudify-cosmo/cloudify-manager#4006

geokala (approved):

PR #1235, opened 2023-01-16
RD-6846 Allow explicit scripts in script_runner

tehasdf:
- not only specialcase `implementation: script/path.py`, but also a script written right there
- set those operations to executor=auto, which means it'll be either host_agent, or CDA, depending on whether it is contained_in a Compute or not

### DIFF
--- a/dsl_parser/elements/operation.py
+++ b/dsl_parser/elements/operation.py
@@ -465,7 +465,7 @@ def _is_inline_script(script):
         return False
 
     if (
-        script.count('.') > 0 and
+        '.' in script and
         all(part.isidentifier() for part in script.split('.'))
     ):
         # this looks like a dotted path - identifiers separated by dots
--- a/dsl_parser/elements/operation.py
+++ b/dsl_parser/elements/operation.py
@@ -312,7 +312,7 @@ def process_operation(
     if is_valid_url or local_resource_exists:
         is_script = False
     else:
-        is_script = _is_explicit_script(operation_mapping)
+        is_script = _is_inline_script(operation_mapping)
 
     if (
         is_valid_url or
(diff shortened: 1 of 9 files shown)


dsl_parser/elements/operation.py
```diff
@@ -415,3 +428,27 @@ def process_operation(
                     'workflow' if is_workflows else 'operation'))
         error_message = base_error_message + partial_error_message
         raise exceptions.DSLParsingLogicException(error_code, error_message)
+
+
+def _is_explicit_script(script):
```
geokala, line 433: Leaving aside the implementation of this (which I'm still considering and we're also discussing out of band), I'd probably call this _is_inline_script

dsl_parser/tests/test_parser_api.py
```diff
@@ -2108,6 +2108,46 @@ def assert_operation(op, extra_properties=False):
         self.assertEqual(workflow2['parameters']['key']['default'], 'value')
         self.assertEqual(workflow['plugin'], constants.SCRIPT_PLUGIN_NAME)
 
+    def test_script_mapping_explicit_script(self):
```
geokala, line 2111: We should probably have a test that covers a single line inline script too?

tehasdf: This test already has one! `op3: "ctx logger info 'hello world'"`

geokala: So it does- nice!

geokala (approved):

dsl_parser/elements/operation.py
```diff
@@ -446,18 +445,25 @@ def _is_inline_script(script):
...
+
+    if '/' in script or '\\' in script:
+        # script is one line and contains a slash/backslash - interpret
+        # this as a path (posix or windows).
+        # This means it's impossible to have 1-line scripts containing
+        # slashes, but this stays backwards-compatible
+        return False
+
+    if (
+        script.count('.') > 0 and
```
geokala, line 468: Isn't this less efficient than "'.' in script"?

geokala (approved):

PR #1238, opened 2023-01-30
add resource_id and is_external properties

EarthmanT:


cloudify/context.py
```diff
@@ -404,7 +404,10 @@ def properties(self):
         These properties are the properties specified in the blueprint.
         """
         self._get_node_if_needed()
-        return self._node.properties
+        try:
+            return self._node.properties
+        except AttributeError:
+            return self._node.get('properties')
```
mateumann, line 410:
```suggestion
        return self._node.get('properties')
```

should do the job

mateumann, line 410: Is it really necessary (other than in test)?

mateumann (commented):
Are you sure you have to include that wagon file?

tehasdf (changes requested):
remember to remove the dev-reqs.txt & .wgn (accidentally committed?)

PR #1250, opened 2023-03-09
RD-6954 rest-client: declare .latest_execution fields

tehasdf:


cloudify_rest_client/deployments.py
```diff
@@ -117,13 +117,6 @@ def deployment_groups(self):
         """
         return self.get('deployment_groups')
 
-    @property
-    def latest_execution_status(self):
```
tehasdf, line 121: just moving this one down to be near the other `latest_execution` fields

mateumann (approved):

PR #1279, opened 2023-05-09
RND-623 Delete dep-group with deployments: return excgroup id

tehasdf:
In order to follow the progress of deleting the deployments, it'll be useful to receive the execution group id


tehasdf:
This is the other part of cloudify-cosmo/cloudify-manager#4211

adrian-polanczyk-codilime (approved):

PR #1322, opened 2024-02-15
NE-19591 Mark not-complete executions as snapshot_cancelled

mateumann:
`snapshot_cancelled` is the new status for executions.  Executions which were still in progress (`pending`, `started`, `cancelling`, `force_cancelling` or `kill_cancelling`) when the snapshot was made, will be marked as `snapshot_cancelled`.  It means a manual examination may be required to determine their state.

Executions with `snapshot_cancelled` status are allowed to be removed.


tehasdf (dismissed):

mateumann:
We will not introduce a new status after all.  This is because we do not want to modify database schema between patch versions (7.0.3 → 7.0.4).

